PR #33, opened 2026-04-10
Remove BYTEA column, add Conversation links to index.md

efenocchi:
### Remove BYTEA content column
All file content is text (markdown, JSONL). The BYTEA `content` column duplicated the `summary` TEXT column with unnecessary hex encoding overhead.
- Remove `content: Buffer` from WriteRow and PendingRow interfaces
- Remove hex encoding from all write paths (upsert, flush, appendFile, session-start, wiki-worker)
- Remove `isText()`, `decodeContent()` helpers and binary mime types
- Prefetch and read paths use `SELECT summary` only
- CREATE TABLE no longer includes `content BYTEA` column

### Add Conversation column to virtual index.md
Direct `[messages]` link to session JSONL files in the index, so Claude can navigate from index to raw conversation data in one step instead of: index → summary → find Source line → construct JSONL path.

### Session files are read-only
All mutating VFS operations (write, append, rm, cp-to, mv) on session JSONL paths now throw EPERM. Previously these either silently created ghost rows in the memory table or faked a delete that reverted on next bootstrap. Reads and cp-from session paths still work.

### Cleanup
- Remove stale compiled .js/.d.ts files from src/ that caused vitest to resolve outdated code
- Fix test import paths

## Test plan

- [x] 386 unit tests passing (vitest)
- [x] E2E real-table-test.mjs: validates text-only schema against live API (33 assertions)
- [x] E2E claude -p --plugin-dir: write, read index.md with Conversation column, follow messages link
- [x] E2E virtual shell: all 6 mutating ops on session paths return EPERM, reads work, normal ops unaffected
- [x] E2E curl verification: writes/deletes hit actual SQL table, no ghost rows


efenocchi:
@claude review again

PR #37, opened 2026-04-11
Remove all FUSE/CLI references from hivemind plugin

kaghni:
Hivemind plugin should be independent from deeplake-cli. Removed CLI detection, FUSE mount checks, install prompts. Unsafe commands now denied with safe alternatives message.


kaghni:
Outdated — FUSE references already removed from main.

PR #38, opened 2026-04-11
Add Codex CLI plugin support

efenocchi:
## Summary

Full Codex CLI plugin implementation alongside the existing Claude Code and OpenClaw plugins. Codex agents can now access the shared Deeplake memory system — reading session summaries, searching across org memory, and capturing their own sessions.

### Codex plugin (`src/hooks/codex/` + `codex/`)
- **SessionStart** — injects memory context as plain text (Codex rejects JSON output)
- **Capture** (UserPromptSubmit + PostToolUse) — writes session events to the sessions table
- **PreToolUse** — intercepts Bash commands targeting the memory path using block+inject strategy (stderr + exit 2). Handles `cat`, `ls`, and `grep` via direct SQL fast path; falls back to virtual shell for other commands
- **Stop** — captures final event and spawns wiki summary worker
- **Wiki worker** — generates structured session summaries using the same template as Claude Code

### VFS improvements
- Suppress "Mounting deeplake://..." stderr messages in one-shot `-c` mode
- Capture stderr in `runVirtualShell()` via `stdio: pipe` instead of inheriting to parent process
- Add `ls` to the codex SQL fast path for directory listings

### Cleanup
- Remove deeplake-cli references from `auth.ts` and hook descriptions
- Remove FUSE/mount language from `CONCERNS.md` and shell docstrings
- Remove `CLAUDE.md` from version control (now local-only via `.gitignore`)
- Revert table names from `memory_test`/`sessions_test` back to `memory`/`sessions`
- Drop the test tables from the DB

### Testing
- New `deeplake-api.test.ts` — covers query retry logic, semaphore concurrency, upsert paths, listTables, ensureTable (26% to 99% statement coverage)
- New codex tests: `codex-hooks.test.ts`, `codex-integration.test.ts`, `codex-capture.test.ts`
- All 443 tests pass, overall coverage 84% stmts / 87% lines

### Build
- `esbuild.config.mjs` now produces 7 codex bundles alongside the 7 claude-code bundles
- Marketplace manifest updated with `hivemind-codex` entry
- README updated with Codex quick start instructions

## Platform differences (Codex vs Claude Code)

| Aspect | Claude Code | Codex |
|--------|------------|-------|
| PreToolUse fires for | Read, Write, Edit, Glob, Grep, Bash | Bash only |
| Hook output format | JSON `updatedInput` | stderr + exit 2 (block+inject) |
| SessionStart output | JSON `additionalContext` | Plain text on stdout |
| SessionEnd | Dedicated hook | Stop does double duty |
| Async hooks | Supported | Not supported |

See `codex/CONCERNS.md` for full details.

## Test plan

- [x] `npm test` — 443 tests pass
- [x] E2E: codex exec with memory queries — reads work, no mount noise
- [x] Smoke test: hook unit tests via piped JSON
- [x] Verified `cat`, `ls`, `ls -la`, `grep` return clean output via SQL fast path
- [x] Verified virtual shell fallback (`head`, `find`, etc.) works without stderr leakage


efenocchi:
@claude review again

kaghni:
## Review — structuring the Codex plugin for the monorepo

Good work getting the hooks working with Codex's different contract (stderr+exit2, plain text output, Bash-only PreToolUse). The platform differences table is helpful. A few things to address before merging:

### 1. Plugin code should live in `codex/src/`, not `src/hooks/codex/`

The monorepo convention is: shared code in `src/`, plugin-specific code in its own dir (`claude-code/src/`, `openclaw/src/`, `codex/src/`). The Codex hooks are currently in `src/hooks/codex/` which pollutes the shared namespace. Move them to `codex/src/hooks/`.

### 2. Don't delete CLAUDE.md

CLAUDE.md has project-wide build instructions and conventions that all contributors need. It should stay in version control.

### 3. Install story — git clone from HEAD is not viable

The `install.sh` does `git clone` + `cp -r` with no version pinning or integrity check. This was flagged twice by the automated review as a supply chain risk.

More importantly: Codex plugins work across CLI, VS Code extension, and Cursor. Users on IDE extensions are not going to run `git clone`. The install script should:
- Pin a specific release tag, not HEAD
- Ideally provide a one-liner that downloads a release artifact

Since Codex has no marketplace yet (self-serve publishing isn't available), we should at minimum make the manual setup as simple as possible — a single config.toml snippet users can paste.

### 4. Reduce duplication — extract shared logic to `src/`

`capture.ts`, `stop.ts`, `session-start.ts`, `wiki-worker.ts`, `pre-tool-use.ts` are near-copies of the CC hooks. The platform-specific parts (output format, hook contract) are real differences, but the shared logic should be extracted:

- Session path building (`buildSessionPath`) — already duplicated in openclaw too
- Wiki summary template
- SQL INSERT patterns for the sessions table
- Safe command detection (`isSafe`)

Each plugin hook becomes a thin adapter: read stdin → call shared logic → format output for the platform. This is the whole point of the monorepo.

### 5. Don't commit 68K-line bundle

`codex/bundle/shell/deeplake-shell.js` is 68,804 lines. Bundle should be built from source, not checked in at that size. If bundles must be checked in (like CC for marketplace), keep it to the hook bundles, not the full shell.

### 6. `agent` column cherry-pick

The `agent` column was added in `feature/codex-implementation` (commit fbdc2fb) and has been cherry-picked to `feat/openclaw-login-command`. Make sure to rebase to avoid conflicts — both branches touch `ensureTable`/`ensureSessionsTable` and capture INSERT statements.

---

**TL;DR**: The hook implementations work, but the code organization needs to follow the monorepo pattern. Plugin-specific code in `codex/src/`, shared logic extracted to `src/`, no git-clone-from-HEAD install, no CLAUDE.md deletion.

.claude-plugin/marketplace.json
```diff
@@ -15,6 +15,13 @@
       "version": "0.6.11",
       "source": "./claude-code",
       "homepage": "https://github.com/activeloopai/hivemind"
+    },
```
kaghni: This will create a new plugin named hivemind-codex on claude code marketplace which doesn't make snece I think

codex/CONCERNS.md
```diff
@@ -0,0 +1,150 @@
+# Codex Plugin — Concerns & Differences from Claude Code
```
kaghni: should be deleted or ignored

codex/install.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
```
kaghni: .sh install should be faster than asking codex to do it with above INSTALL.md and we hsould also be able to directly ask codex to run the install.sh . SO let's either keep the install.sh or INSTALL.md

codex/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "name": "hivemind-codex",
+  "version": "0.6.7",
```
kaghni: The versions are not matching with the global pakcage.json the update logic for codex should be added to release.yaml so they'll stay synced with other plugins

src/hooks/codex/capture.ts
```diff
@@ -0,0 +1,118 @@
+#!/usr/bin/env node
```
kaghni, line 1: Later should be moved to codex/hooks or shared if these are shared between codex claude and openclaw. but we cna leave it for now

src/hooks/codex/pre-tool-use.ts
```diff
@@ -0,0 +1,240 @@
...
+  "md5sum", "sha1sum", "sha256sum",
+  "echo", "printf", "tee",
+  "pwd", "cd", "basename", "dirname", "env", "printenv", "hostname", "whoami",
+  "date", "seq", "expr", "sleep", "timeout", "time", "true", "false", "test",
+  "alias", "unalias", "history", "help", "clear",
+  "for", "while", "do", "done", "if", "then", "else", "fi", "case", "esac",
+]);
+
+function isSafe(cmd: string): boolean {
+  const stripped = cmd.replace(/'[^']*'/g, "''").replace(/"[^"]*"/g, '""');
```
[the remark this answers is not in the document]
kaghni: can you double check this

kaghni (changes requested):

PR #40, opened 2026-04-13
fix: improve CLI handoff and fix auto-update version discovery

efenocchi:
## Summary

- **Soft guidance instead of hard deny**: When agents try unsupported commands (python, node, curl) targeting the memory filesystem, the pre-tool-use hook now returns actionable guidance ("use cat, jq, grep instead") rather than a `permissionDecision: deny`. This eliminates the red-line alarm loops where agents either retry the same blocked command or give up entirely.

- **System prompt steers agents to bash**: The session-start injection now explicitly tells agents that only bash builtins are available for memory operations — no python, node, or curl. This prevents the problem proactively before agents even attempt an unsupported command.

- **Auto-update version discovery fix**: `getInstalledVersion()` was looking for `package.json` exactly one level up from the bundle directory, which doesn't exist in either the cache or marketplace install layouts. It now walks up to 5 parent directories, matching on `name: "hivemind"`. This was the root cause of auto-update never triggering for installed plugins (chicken-and-egg: the old version couldn't find its own version to know it was outdated).

- **Deeplake CLI passthrough**: Commands like `deeplake mount`, `deeplake login`, `deeplake status` now pass through to the real shell instead of being blocked.

## Changes

| File | What |
|------|------|
| `src/hooks/session-start.ts` | Add "no python/node/curl" to prompt; walk-up `getInstalledVersion` |
| `src/hooks/codex/session-start.ts` | Same for Codex |
| `src/hooks/pre-tool-use.ts` | Soft guidance + CLI passthrough |
| `src/hooks/codex/pre-tool-use.ts` | Same for Codex |
| `claude-code/tests/pre-tool-use.test.ts` | **New** — 30 tests for hook behavior |
| `claude-code/tests/autoupdate.test.ts` | 6 new tests for version walk-up |
| `codex/tests/codex-integration.test.ts` | 5 new edge case tests |

## Test plan

- [x] All 486 tests pass (23 test files)
- [x] `claude -p "list files in deeplake memory" --plugin-dir ./claude-code` — works, agent uses ls/cat
- [x] `claude -p "count keys in JSON files in deeplake memory. Use cat and jq" --plugin-dir ./claude-code` — works, agent uses bash+jq
- [x] Unsafe commands (python3, node, curl) return guidance message, not deny
- [x] Deeplake CLI commands (mount, login) pass through unmodified
- [x] Direct hook test: `echo '{"tool_name":"Bash","tool_input":{"command":"ls ~/.deeplake/memory/"},...}' | node bundle/pre-tool-use.js` returns allow with shell rewrite

### DIFF
--- a/claude-code/bundle/session-start.js
+++ b/claude-code/bundle/session-start.js
@@ -333,19 +333,29 @@ Organization management \u2014 each argument is SEPARATE (do NOT quote subcomman
 - node "DEEPLAKE_AUTH_CMD" members                            \u2014 list members
 - node "DEEPLAKE_AUTH_CMD" remove <user-id>                   \u2014 remove member
 
+IMPORTANT: Only use bash commands (cat, ls, grep, echo, jq, head, tail, etc.) to interact with ~/.deeplake/memory/. Do NOT use python, python3, node, curl, or other interpreters \u2014 they are not available in the memory filesystem. If a task seems to require Python, rewrite it using bash commands and standard text-processing tools (awk, sed, jq, grep, etc.).
+
 LIMITS: Do NOT spawn subagents to read deeplake memory. If a file returns empty after 2 attempts, skip it and move on. Report what you found rather than exhaustively retrying.
 
 Debugging: Set DEEPLAKE_DEBUG=1 to enable verbose logging to ~/.deeplake/hook-debug.log`;
 var GITHUB_RAW_PKG = "https://raw.githubusercontent.com/activeloopai/hivemind/main/package.json";
 var VERSION_CHECK_TIMEOUT = 3e3;
 function getInstalledVersion() {
-  try {
-    const pkgPath = join4(__bundleDir, "..", "package.json");
-    const pkg = JSON.parse(readFileSync3(pkgPath, "utf-8"));
-    return pkg.version ?? null;
-  } catch {
-    return null;
+  let dir = __bundleDir;
+  for (let i = 0; i < 5; i++) {
+    const candidate = join4(dir, "package.json");
+    try {
+      const pkg = JSON.parse(readFileSync3(candidate, "utf-8"));
+      if (pkg.name === "hivemind" && pkg.version)
+        return pkg.version;
+    } catch {
+    }
+    const parent = dirname(dir);
+    if (parent === dir)
+      break;
+    dir = parent;
   }
+  return null;
 }
 async function getLatestVersion() {
   try {
--- a/claude-code/bundle/session-start.js
+++ b/claude-code/bundle/session-start.js
@@ -346,7 +346,7 @@ function getInstalledVersion() {
     const candidate = join4(dir, "package.json");
     try {
       const pkg = JSON.parse(readFileSync3(candidate, "utf-8"));
-      if (pkg.name === "hivemind" && pkg.version)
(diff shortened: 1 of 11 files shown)


codex/bundle/pre-tool-use.js
```diff
@@ -431,8 +427,14 @@ async function main() {
     return;
   const rewritten = rewritePaths(cmd);
   if (!isSafe(rewritten)) {
-    log3(`unsafe command blocked: ${rewritten}`);
-    blockWithDeny("This command is not supported for memory operations. Use standard commands like cat, ls, grep, echo instead.");
+    if (/^\s*deeplake\s+(mount|login|unmount|status)\b/.test(cmd) && !/[;&|]/.test(cmd) || cmd.includes("deeplake.ai/install")) {
+      log3(`deeplake CLI command \u2014 passing through`);
```
kaghni, line 431: Since we decide to remove CLI connection. Do we need this part?

src/hooks/codex/pre-tool-use.ts
```diff
@@ -130,8 +124,21 @@ async function main(): Promise<void> {
   const rewritten = rewritePaths(cmd);
 
   if (!isSafe(rewritten)) {
-    log(`unsafe command blocked: ${rewritten}`);
-    blockWithDeny("This command is not supported for memory operations. Use standard commands like cat, ls, grep, echo instead.");
+    // Pass through deeplake CLI commands (mount, login, etc.)
+    // Anchored to start of string and reject chained commands (;, &&, ||, |).
+    if (/^\s*deeplake\s+(mount|login|unmount|status)\b/.test(cmd) && !/[;&|]/.test(cmd) || cmd.includes("deeplake.ai/install")) {
+      log(`deeplake CLI command — passing through`);
```
kaghni, line 130: Same here

src/hooks/pre-tool-use.ts
```diff
@@ -135,13 +135,34 @@ async function main(): Promise<void> {
...
   const shellCmd = getShellCommand(input.tool_name, input.tool_input);
 
-  if (!shellCmd && touchesMemory(cmd)) {
-    log(`unsafe command blocked: ${cmd}`);
+  // Also check non-Bash tools (Read/Write/Edit/Glob/Grep) that touch memory but didn't get a shellCmd
+  const toolPath = (input.tool_input.file_path ?? input.tool_input.path ?? "") as string;
+  if (!shellCmd && (touchesMemory(cmd) || touchesMemory(toolPath))) {
+    // Pass through deeplake CLI commands (mount, login, etc.) to real bash.
+    // Anchored to start of string and reject chained commands (;, &&, ||, |).
+    if (/^\s*deeplake\s+(mount|login|unmount|status)\b/.test(cmd) && !/[;&|]/.test(cmd) || cmd.includes("deeplake.ai/install")) {
```
kaghni, line 143: Ah ok so we still getting somehow agent triggering deeplake cli command and this way we try to avoid them? But why are we still getting them after removing all deeplake cli connections

claude-code/bundle/session-start.js
kaghni, line 316: Can you also add same session-star changes to SKILl.md? To keep them in sync so throughout the session if needed agent will load the SKILL.md

kaghni (changes requested):

PR #55, opened 2026-04-16
Feat/locomo accuracy improvements

kaghni:
## Version Bump

> **To trigger a release**, bump `"version"` in `package.json` before merging.
>
> | Change type     | Version bump          | Example              |
> | --------------- | --------------------- | -------------------- |
> | Bug fix         | patch (1.2.0 → 1.2.1) | `"version": "1.2.1"` |
> | New feature     | minor (1.2.0 → 1.3.0) | `"version": "1.3.0"` |
> | Breaking change | major (1.2.0 → 2.0.0) | `"version": "2.0.0"` |
>
> If you don't bump the version, no release will be created.

## Test plan

- [ ] Tests pass locally (`npm test`)
- [ ] Relevant new tests added
- [ ] Version bumped in `package.json`, or no release needed for this change

PR #67, opened 2026-04-21
feat: LoCoMo accuracy improvements — auto-read + multi-word OR

kaghni:
## Summary

Two focused improvements to memory retrieval accuracy, benchmarked against the [LoCoMo](https://github.com/snap-research/locomo) conversational memory benchmark on a balanced 60-question subset (15 per category × 4 non-adversarial categories, drawn across all 10 conversations).

## Benchmark results

Model: `claude -p --model haiku`. Judge: `openai/gpt-5-mini`. See [deeplake-blitz PR #1](https://github.com/activeloopai/deeplake-blitz/pull/1) for methodology and raw results.

| Config | Overall | single-hop | temporal | multi-hop | open-domain |
|---|---|---|---|---|---|
| `origin/main` (this branch's base) | **46.7%** | — | — | — | — |
| + auto-read (`64f2246`) | **58.3%** | — | — | — | — |
| + multi-word OR (`04bf1d5`, **this PR**) | **59.2%** | — | — | — | — |

Net: **+12.5% absolute** over `origin/main` on a category-balanced 60-question sample.

All 918 existing tests pass.

### 1. `feat(pre-tool-use): auto-rewrite unsupported interpreter reads to cat` (`64f2246`)

When the agent invokes an unsupported interpreter (`python`/`python3`/`node`/`deno`/`bun`/`ruby`/`perl`) on a single memory path with no shell metacharacters, rewrite the call to `cat '<path>'` instead of returning the generic `[RETRY REQUIRED]` guidance. Avoids burning a turn when the intent is clearly a plain file read.

Narrow safety conditions:
- Command starts with a recognized read-like interpreter.
- No shell metacharacters (`$`, `` ` ``, `;`, `|`, `&`, `<`, `>`, `()`, `\`).
- Memory path matches conservative `[\w./_-]+` regex (no trailing quotes/parens can glue onto it).
- Final `cat` command wraps the path in single quotes with `'\''`-escape.

Composite commands (curl, wget, chains, pipes, substitutions) still fall through to the RETRY guidance.

### 2. `feat(grep-core): multi-word OR split for non-regex patterns` (`04bf1d5`)

When a non-regex grep pattern contains multiple words, split into per-word OR filters instead of requiring the full literal phrase to appear verbatim. Natural-language queries like `Melanie kids dinosaurs` now match sessions that mention any of the three tokens, not only those that contain all three in that exact order.

Single-word and regex patterns are unchanged.

Adds an optional `multiWordPatterns` field on `SearchOptions`. `buildGrepSearchOptions` populates it when the pattern splits into >1 word-long tokens. `searchDeeplakeTables` consumes it in the non-contentScanOnly branch, falling back to `escapedPattern`. All downstream OR joining is handled by the existing `buildContentFilter` helper.

## What was considered and dropped

During development, a third change — prepending extracted `date_time` to the first grep match per file — was committed and tested. It **regressed** from 59.2% → 50.8%, so it was removed from this PR. Session date extraction may still be a valid idea in a different form (e.g. as a separate structured field instead of inline prefix), but the naive prefix hurts more than it helps on this benchmark.

A previous branch (closed PR #55) had 9 commits based on the pre-refactor codebase. Only these two improvements survived the port to `main`'s new `grep-core.ts`/`pre-tool-use.ts` architecture and actually improved accuracy under proper benchmarking conditions. The rest were either obsolete (already done by `grep-core.ts` UNION ALL, PR #64's index fix) or didn't carry over cleanly.

## Methodology note

Early benchmark runs (on PR #55) had two bugs that inflated apparent gains:
1. Benchmark repo checked out to wrong branch, so `--subset` flag was silently ignored and benchmark ran all 1540 questions with many timeouts → 0-5% false "results".
2. Benchmark script set `DEEPLAKE_CAPTURE=false` but `main`'s new code reads `HIVEMIND_CAPTURE`, so session writes leaked into `locomo_sessions` mid-run, polluting search results.

Both fixes live in the deeplake-blitz PR. All numbers above are post-fix.

## Test plan
- [x] `npm test` passes (918/918 tests, including 32 pre-tool-use.test.ts tests that specifically cover the RETRY-vs-cat rewrite boundary for curl, command substitution, backticks, interpreter commands, etc.)
- [x] `npm run build` produces matching bundles
- [x] Balanced 60q LoCoMo benchmark: +12.5% over `main`

### DIFF
--- a/claude-code/bundle/pre-tool-use.js
+++ b/claude-code/bundle/pre-tool-use.js
@@ -639,9 +639,9 @@ function buildPathCondition(targetPath) {
   return `(path = '${sqlStr(clean)}' OR path LIKE '${sqlLike(clean)}/%' ESCAPE '\\')`;
 }
 async function searchDeeplakeTables(api, memoryTable, sessionsTable, opts) {
-  const { pathFilter, contentScanOnly, likeOp, escapedPattern, prefilterPattern, prefilterPatterns } = opts;
+  const { pathFilter, contentScanOnly, likeOp, escapedPattern, prefilterPattern, prefilterPatterns, multiWordPatterns } = opts;
   const limit = opts.limit ?? 100;
-  const filterPatterns = contentScanOnly ? prefilterPatterns && prefilterPatterns.length > 0 ? prefilterPatterns : prefilterPattern ? [prefilterPattern] : [] : [escapedPattern];
+  const filterPatterns = contentScanOnly ? prefilterPatterns && prefilterPatterns.length > 0 ? prefilterPatterns : prefilterPattern ? [prefilterPattern] : [] : multiWordPatterns && multiWordPatterns.length > 1 ? multiWordPatterns : [escapedPattern];
   const memFilter = buildContentFilter("summary::text", likeOp, filterPatterns);
   const sessFilter = buildContentFilter("message::text", likeOp, filterPatterns);
   const memQuery = `SELECT path, summary::text AS content, 0 AS source_order, '' AS creation_date FROM "${memoryTable}" WHERE 1=1${pathFilter}${memFilter} LIMIT ${limit}`;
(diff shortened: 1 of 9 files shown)


efenocchi:
LGTM

efenocchi (approved):

PR #95, opened 2026-05-06
fix(auth-login): validate workspace on switch; reset stale workspace on org switch

### LINKED ISSUE #92: workspace switching doesn't validate existence; org switch silently leaves stale workspace
## Summary

Two related bugs in the `auth-login` CLI around org/workspace switching:

1. **`hivemind org switch <org>` doesn't reset/revalidate the saved `workspaceId`.** If you were on workspace `X` in org A, then switch to org B which has no workspace `X`, the saved credentials still point to `workspaceId: X`. Subsequent commands then run against a workspace that doesn't exist in the new org, with no warning.

2. **`hivemind workspace <id>` accepts any string without verifying the workspace exists.** It just writes whatever you typed to credentials. e.g. `hivemind workspace k7` succeeds and prints `Switched to workspace: k7` even when no workspace `k7` exists in the current org.

Combined, this means a user who types `hivemind org switch activeloop` (forgetting to also fix the workspace) ends up silently misconfigured, and even if they try to fix it with `hivemind workspace <id>` there's no feedback that the id is wrong.

## Repro

```bash
# starting state: org A, workspace foo (where foo exists in A)
hivemind org switch activeloop
hivemind whoami
# -> still shows Workspace: foo, even though `foo` doesn't exist in activeloop

hivemind workspace k7
# -> "Switched to workspace: k7" with no validation, even if k7 isn't a real workspace
```

## Source

- `src/commands/auth.ts:166-170` — `switchOrg` saves `{ orgId, orgName }` without touching `workspaceId`.
- `src/commands/auth.ts:180-184` — `switchWorkspace` calls `saveCredentials({ ...creds, workspaceId })` with no existence check.
- `src/commands/auth-login.ts:83-90` — CLI dispatch for `workspace <id>` doesn't call `listWorkspaces` to verify.

Compare to the `org switch` CLI handler at `src/commands/auth-login.ts:62-69`, which *does* call `listOrgs` and rejects unknown names — the workspace path should mirror that.

## Suggested fix

- In the `workspace <id>` CLI handler, call `listWorkspaces(creds.token, apiUrl, creds.orgId)` first and reject if the id/name isn't found (same pattern as `org switch`).
- In `switchOrg` (or its CLI handler), after switching, check whether the existing `workspaceId` still exists in the new org's workspace list. If not, either:
  - reset to `"default"` (matching the post-login default), or
  - print a warning and prompt the user to pick one via `hivemind workspaces` + `hivemind workspace <id>`.

Either behavior is better than silently keeping a stale workspace.

## Environment

- hivemind 0.6.46 (also visible in 0.7.4 from a quick read of the same file)

efenocchi:
Closes #92.

## Summary

- `hivemind workspace <id>` no longer silently saves an unknown id. It now lists workspaces, validates id-or-name (case-insensitive), and exits 1 with the available list on miss — same UX as `hivemind org switch`.
- `hivemind org switch` now resets the active workspace to `default` (with a warning) when the previously-saved workspace doesn't exist in the new org. Previously the saved `workspaceId` carried over silently, leaving the user pointing at a workspace that didn't exist in their new org.
- The `workspace` command surface is restructured into `workspace list` and `workspace switch <name-or-id>`, mirroring `org list` / `org switch`. The bare-target shortcut (`workspace <id>`) is removed.

## Why

Two related bugs in the auth-login CLI dispatcher:

1. `switchWorkspace` accepted any string and persisted it to credentials with no existence check. e.g. `hivemind workspace k7` printed `Switched to workspace: k7` even when `k7` was not a real workspace in the current org.
2. `switchOrg` saved `{ orgId, orgName }` without touching `workspaceId`. After `hivemind org switch <new-org>`, you were still scoped to the previous workspace id — which often didn't exist in the new org, so subsequent commands ran against a phantom workspace.

Combined, a user who typed `hivemind org switch <other-org>` (forgetting to also fix the workspace) ended up silently misconfigured, and even an attempt to fix it with `hivemind workspace <id>` produced no feedback that the id was wrong.

## What changed

**`src/commands/auth-login.ts`**
- `org switch` handler: after the switch, calls `listWorkspaces(creds.token, apiUrl, <new-org-id>)` and, if the carried-over `workspaceId` isn't in the result, calls `switchWorkspace("default")` and prints a warning + the available list.
- `workspace` handler: split into two subcommands.
  - `workspace list` mirrors the existing top-level `workspaces` command.
  - `workspace switch <target>` does the validate-then-switch, exiting 1 with the available list on miss.
  - Anything else under `workspace` (including the previously-supported bare target) is rejected with usage.
- Docstring updated.

**`src/cli/index.ts`**
- `hivemind --help` now documents `workspace list` and `workspace switch <name-or-id>` instead of the bare `workspace <id>` form.

**`claude-code/tests/auth-login-dispatch.test.ts`**
- Existing `org switch` tests updated to mock the new `listWorkspaces` call.
- Added 10 cases covering:
  - `workspace switch <unknown>` exits 1 with the available list, never calls `switchWorkspace`.
  - `workspace switch <id>` and `workspace switch <name>` (case-insensitive) both resolve correctly.
  - `workspace switch` (no target) exits 1 without hitting the API.
  - `workspace list` is read-only — never calls `switchWorkspace`.
  - `workspace <target>` (the removed bare form) is rejected without API calls — explicit regression guard.
  - `org switch` resets the workspace to `default` when the carried-over id isn't in the new org's list.
  - `org switch` does NOT reset the workspace when the carried-over id (or name) IS still present in the new org.
  - `listWorkspaces` is called with the NEW org id, not the stale one from `creds`.

**Bundles** (`bundle/cli.js`, `claude-code/bundle/commands/auth-login.js`, `codex/bundle/commands/auth-login.js`, `cursor/bundle/commands/auth-login.js`, `hermes/bundle/commands/auth-login.js`)
- Rebuilt via `npm run build`. Second commit is a pure rebuild — no source edits.

## Test plan

- [x] `npx vitest run claude-code/tests/auth-login-dispatch.test.ts` → 42/42 pass
- [x] `npm run build` → all bundles rebuild cleanly
- [x] Local install + smoke tests (overlaid `bundle/cli.js` onto the global `@deeplake/hivemind` install):
  - `hivemind workspace list` → lists current org's workspaces
  - `hivemind workspace switch <real-name>` → switches successfully
  - `hivemind workspace switch <bogus>` → exits 1 with available list, no state mutation
  - `hivemind workspace switch` (no arg) → exits 1 with subcommand usage, no API call
  - `hivemind workspace <bare-name>` → exits 1 with the new usage line
  - `hivemind workspace` (no args) → exits 1 with the new usage line
  - `hivemind whoami` after each → confirms credentials state is consistent with the last successful switch
- [ ] Reviewer to verify in a multi-org account: `hivemind org switch <other-org>` from a workspace not present in the target org → resets to `default` with the warning message. (Needs an account with at least two orgs to repro; the unit tests cover the dispatcher logic deterministically.)

## Breaking change

`hivemind workspace <id>` (the bare-target shortcut) no longer switches; it now exits 1 with usage. Replace with `hivemind workspace switch <id>`. The plural `hivemind workspaces` command and the `/hivemind_switch_workspace` OpenClaw slash command are unaffected.

## Release Notes

* **New Features**
  * Added `workspace list` and `workspace switch <name-or-id>` subcommands for improved workspace management.
  * Implemented automatic workspace validation during organization switches; resets to default if the workspace is unavailable in the new organization.

* **Documentation**
  * Updated CLI usage documentation to reflect new workspace management subcommands.

### DIFF
--- a/src/commands/auth-login.ts
+++ b/src/commands/auth-login.ts
@@ -110,7 +110,7 @@ export async function runAuthCommand(args: string[]): Promise<void> {
     case "workspaces": {
       if (!creds) { console.log("Not logged in."); process.exit(1); }
       const ws = await listWorkspaces(creds.token, apiUrl, creds.orgId);
-      ws.forEach(w => console.log(`${w.id}  ${w.name}`));
+      ws.forEach(w => console.log(w.name || w.id));
       break;
     }
 
@@ -120,7 +120,7 @@ export async function runAuthCommand(args: string[]): Promise<void> {
 
       if (sub === "list") {
         const wsList = await listWorkspaces(creds.token, apiUrl, creds.orgId);
-        wsList.forEach(w => console.log(`${w.id}  ${w.name}`));
+        wsList.forEach(w => console.log(w.name || w.id));
         break;
       }
 
--- a/src/commands/auth-login.ts
+++ b/src/commands/auth-login.ts
@@ -66,21 +66,40 @@ export async function runAuthCommand(args: string[]): Promise<void> {
         const orgs = await listOrgs(creds.token, apiUrl);
         const match = orgs.find(o => o.id === target || o.name.toLowerCase() === target.toLowerCase());
         if (!match) { console.log(`Org not found: ${target}`); process.exit(1); }
+
+        // Resolve the carry-over BEFORE mutating credentials. If listWorkspaces
+        // fails (network blip, 5xx) the exception bubbles up unchanged and the
+        // org switch never happens — re-running the command then succeeds
+        // cleanly instead of leaving credentials half-committed.
+        const prevWs = creds.workspaceId ?? "default";
+        const lcPrev = prevWs.toLowerCase();
+        const wsList = await listWorkspaces(creds.token, apiUrl, match.id);
+        // Resolve to the matched workspace OBJECT, not a boolean: `workspaceId`
+        // is supposed to be a canonical id but legacy creds (and the post-login
+        // `"default"` sentinel) can hold a name. We need the matched object so
+        // we can normalize a name-only match to the canonical id.
+        const matchedWs = wsList.find(w => w.id === prevWs || (w.name && w.name.toLowerCase() === lcPrev));
+
         await switchOrg(match.id, match.name);
         console.log(`Switched to org: ${match.name}`);
 
-        // Carry-over guard: the previous workspaceId may not exist in the new
-        // org. If it doesn't, reset to "default" so subsequent commands don't
-        // silently target a non-existent workspace.
-        const prevWs = creds.workspaceId ?? "default";
-        const wsList = await listWorkspaces(creds.token, apiUrl, match.id);
-        const stillThere = wsList.some(w => w.id === prevWs || w.name.toLowerCase() === prevWs.toLowerCase());
-        if (!stillThere) {
-          await switchWorkspace("default");
-          console.log(`Workspace '${prevWs}' is not in org '${match.name}'. Reset workspace to 'default'.`);
-          if (wsList.length > 0) {
-            console.log(`Available workspaces: ${wsList.map(w => w.name || w.id).join(", ")}`);
+        if (!matchedWs) {
+          // Suppress the reset-and-warn when prevWs is already the "default"
+          // sentinel and the new org has no workspace named/idd "default" —
+          // there's nothing to reset and the warning would be misleading.
+          if (prevWs !== "default") {
+            await switchWorkspace("default");
+            console.log(`Workspace '${prevWs}' is not in org '${match.name}'. Reset workspace to 'default'.`);
+            if (wsList.length > 0) {
+              console.log(`Available workspaces: ${wsList.map(w => w.name || w.id).join(", ")}`);
+            }
           }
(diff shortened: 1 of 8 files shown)


src/commands/auth-login.ts
```diff
@@ -65,8 +66,41 @@ export async function runAuthCommand(args: string[]): Promise<void> {
...
+
+        if (!matchedWs) {
+          // Suppress the reset-and-warn when prevWs is already the "default"
+          // sentinel and the new org has no workspace named/idd "default" —
+          // there's nothing to reset and the warning would be misleading.
+          if (prevWs !== "default") {
+            await switchWorkspace("default");
+            console.log(`Workspace '${prevWs}' is not in org '${match.name}'. Reset workspace to 'default'.`);
+            if (wsList.length > 0) {
+              console.log(`Available workspaces: ${wsList.map(w => w.name || w.id).join(", ")}`);
```
kaghni, line 94: listing as a  column would be better. instead of row seperated with commas

efenocchi: solved in a different way

kaghni (approved):

PR #96, opened 2026-05-06
feat(notifications): centralized push-notification framework + Claude Code adapter

kaghni:
## Summary

A trigger-agnostic notification framework that surfaces server-pushed and locally-derived messages to users at session start.

Three streams feed a single `drainSessionStart`:
- **Pull rules** — declarative, evaluated on every drain (e.g. `welcomeRule` fires once per fresh `creds.savedAt`).
- **Push queue** — file-backed at `~/.deeplake/notifications-queue.json`; any code path can `enqueueNotification(...)` and it surfaces at the next session.
- **Backend fetch** — `GET /me/notifications` on deeplake-api with a 1.5s timeout, fail-soft (network error / non-200 / malformed JSON degrade to "no backend notifications").

## Per-agent delivery — NOT uniform

Phase 0 deep-read of each agent's harness source revealed the multi-hook pattern that works on Claude Code does NOT work on Codex/Cursor/Hermes:

| Agent | Multi-hook → distinct context blocks? | v1 delivery |
|---|---|---|
| Claude Code | ✅ separate `<system-reminder>` per command | **REAL** — second SessionStart hook command |
| Codex | ❌ `flatten_additional_contexts` joins with `\n\n` | STUB no-op + TODO |
| Hermes | ❌ `on_session_start` return value DISCARDED at `run_agent.py:9777-9786` (latent upstream bug) | STUB no-op + TODO |
| Cursor | ⚠️ closed-source; docs imply concat | STUB no-op + TODO |

Source-level evidence in `src/notifications/AGENT_CHANNELS.md`. Real Codex/Cursor/Hermes delivery is deferred to a follow-up — likely inline-append for Codex/Cursor and `pre_llm_call` + per-session dedup for Hermes.

## Persistence + safety

- `~/.deeplake/notifications-state.json` — dedup state, atomic tmp+rename writes, self-guards against writes outside HOME (per the 2026-04-28 destructive-command post-mortem rules).
- `~/.deeplake/notifications-queue.json` — same atomicity + sandbox guard.
- Server IDs are prefixed with `backend:` to avoid collision with locally-issued ids.

## Tests

- 28 new tests in `claude-code/tests/notifications.test.ts` covering unit (rules, dedup, format anti-pattern), bundle smoke (testing rule #11), cross-instance race, and backend fetch with mocked global `fetch` (200/500/malformed-JSON/no-creds/dropped-malformed-entries).
- Updated `claude-code/tests/session-start.test.ts` for the 3-hook SessionStart shape.
- Full suite: 1837 pass, 0 fail. `npm run ci` green.

## Companion PR

The backend endpoint lives in **activeloopai/deeplake-api#TBD** — frontend code degrades gracefully (welcome only) until the backend ships.

## Test plan

- [x] `npm test` — 1837 pass
- [x] `npm run dup` clean
- [x] Bundled hook smoke-tested against a mock backend at 127.0.0.1:8765 — three notifications (welcome + 2 server-pushed) rendered into a single `additionalContext` block, second run dedup'd to empty stdout
- [x] Cache overlay test — fresh session in active Claude Code surfaces the welcome
- [ ] Re-run e2e against the real `feat/notifications-endpoint` deeplake-api branch (in progress as part of this work)
- [ ] Codex/Cursor/Hermes follow-up PRs (deferred)

## Summary by CodeRabbit

* **New Features**
  - Session notifications system that displays informational messages and important updates when starting sessions in Claude Code
  - Automatic duplicate prevention ensures notifications appear only once
  - Server-side notifications are fetched and delivered during session initialization
  - Welcome notification provides context when credentials are available

* **Tests**
  - Comprehensive test coverage for notification delivery, queue management, and state persistence

* **Documentation**
  - Added documentation on notification delivery across supported editors


kaghni:
Followups + v1.5 work tracked in #99 (single rolling issue).

kaghni:
## Real-user QA: 19 / 20 ✅

Walked the full Tier 1–4 matrix against the live `feat/notifications-endpoint` deeplake-api branch (real Postgres-18 + real JWT + real httpapi.Server bound to `:8080`). Every case ran as a fresh `claude` session in a real terminal, no synthetic stdin substitutes. New design tweak in 3093d50 corrects the delivery mechanism after empirical probing — see commit message.

### Tier 1 — golden paths (blocking): 5/5 ✅

| # | Scenario | Outcome |
|---|---|---|
| T1.1 | Fresh login, 2 backend notifications | ✅ 3 banners visible verbatim at session start |
| T1.2 | Same-session reload (dedup) | ✅ second fresh `claude` shows nothing — dedup state suppresses |
| T1.3 | Backend with 0 rows | ✅ welcome only, no empty block, no error |
| T1.4 | Backend unreachable | ✅ welcome only; ECONNREFUSED returns in ~2ms (instant), 1.5s timeout never had to fire |
| T1.5 | Re-login refires welcome | ✅ bumping `creds.savedAt` causes welcome to re-emit; backend rows fire too (different UUIDs) |

### Tier 2 — content variation: 5/5 ✅

| # | Scenario | Outcome |
|---|---|---|
| T2.1 | 4000-char body | ✅ renders fully (3900 chars), no truncation. UX flag raised: too long for a banner — see #99 |
| T2.2 | Markdown / code blocks / unicode in body | ✅ bash code blocks, `**bold**`, special chars `< > & " ' \ /`, multilingual (CJK/Cyrillic/Arabic/Devanagari), math `∑ ∫ √ π ≠`, literal newlines all preserved. Caveat: ZWJ emoji clusters (`👨‍👩‍👧‍👦`, `🏳️‍🌈`) get mangled by the terminal renderer — not our code; see #99 |
| T2.3 | All three severities at once | ✅ 🐝 (info) / ⚠️ (warn) / 🚨 (error) prefixes distinct on screen |
| T2.4 | 5+ pending notifications | ✅ 7 backend + welcome = 8 banners render cleanly, no truncation |
| T2.5 | Org-targeted notification | ✅ direct user-target + org-fanout indistinguishable to user; `WHERE target_user_id = $1 OR target_org_id = ANY($2)` plus org-membership join works through real auth+DB stack |

### Tier 3 — concurrency + state: 3/4 ✅, 1 known FAIL

| # | Scenario | Outcome |
|---|---|---|
| T3.1 | Two concurrent sessions cross-instance race | ❌ FAIL — programmatic 3-way concurrent invoke produces 3 duplicate emissions. State file integrity holds (atomic tmp+rename), but at-most-once delivery is not guaranteed under race. **Deferred to v1.5** (see #99 for fix paths) — failure mode is "duplicate banner on rare ms-scale race", not data corruption. |
| T3.2 | `/clear` mid-session | ✅ CC doesn't refire SessionStart on `/clear` |
| T3.3 | `/resume` of recent session | ✅ CC re-fires hook, but our state-file dedup correctly suppresses re-emission. No duplicates visible to user. |
| T3.4 | Task / subagent spawn | ✅ CC 2.1.131 doesn't fire SessionStart hooks for Task-tool subagents. No duplicate banner. |

### Tier 4 — failure observation: 6/6 ✅

| # | Scenario | Outcome |
|---|---|---|
| T4.1 | Expired/invalid JWT (401) | ✅ welcome only, fail-soft. Debug: `fetch ... returned 401` |
| T4.2 | Backend 500 | ✅ welcome only, fail-soft. Debug: `fetch ... returned 500` |
| T4.3 | Malformed JSON body | ✅ welcome only, JSON.parse caught. Debug: `Unexpected token 'h' ... is not valid JSON` |
| T4.4 | Slow backend (3s, our timeout 1.5s) | ✅ AbortController fires at 1.5s; wall-clock 1536ms (Node startup + abort budget). Welcome only. Debug: `This operation was aborted` (the exact DOMException string) |
| T4.5 | Corrupt `notifications-state.json` | ✅ self-healing — corrupt state treated as empty, full 3 banners deliver, file rewritten with valid JSON |
| T4.6 | `~/.deeplake/` not writable (chmod 555) | ✅ banners emit successfully, EACCES on write caught and logged. Trade-off: state doesn't persist across sessions when dir is read-only. Acceptable v1 behavior. |

### Latency profile

Direct bundle invocation timing (real launcher backend at `:8080`):
- Node.js cold-start: ~30ms
- Local I/O (read state + queue): 1-3ms
- Backend HTTP fetch (localhost): ~36ms
- Emit + state write: <1ms
- **Total: ~70ms**

Per-hook breakdown from a real `claude` session: our hook (`session-notifications.js`) runs in **43ms** total. Both hooks (memory + ours) fire **in parallel**, so we don't add to the critical path. The 3-5s perceived session startup the user observed was almost entirely the existing `session-start.js` doing autoupdate (~2.2s) — not new with this PR. (See #99 for the autoupdate-on-every-session followup.)

### What's tracked, what's not

- All followups consolidated in **#99** (single rolling issue) — body length cap, ZWJ rendering caveat, T3.1 race fix paths, admin tooling sketch, per-agent expansion (Codex/Cursor/Hermes/Pi)
- Cross-linked from the deeplake-api PR #209
- Plan file with full QA narrative at `/home/ubuntu/.claude/plans/cool-darling-you-re-gonna-goofy-garden.md` (local-only, retained for audit)

### Confidence

95% — every layer empirically validated against a real Claude Code 2.1.131 session, real Go server, real Postgres-18, real JWT auth. Untested in this round: production deployment of deeplake-api with the new endpoint (waiting on #209 to merge); CC versions older than 2.1.131 (`systemMessage` field is 2.0+ but rendering varies — VS Code extension doesn't render it per anthropics#15344).

src/notifications/queue.ts
```diff
@@ -0,0 +1,56 @@
...
+  const tmp = `${path}.${process.pid}.tmp`;
+  writeFileSync(tmp, JSON.stringify(q, null, 2), { mode: 0o600 });
+  renameSync(tmp, path);
+}
+
+/** Append a notification to the persistent queue. Cross-process safe. */
+export function enqueueNotification(n: Notification): void {
+  const q = readQueue();
+  q.queue.push(n);
+  writeQueue(q);
```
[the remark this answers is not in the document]
kaghni: Acknowledged. The race here (concurrent producers losing updates on `enqueueNotification`) is the same family as the cross-instance race we hit empirically in T3.1 of the QA matrix (concurrent SessionStart hooks → `state.ts` losing dedup writes).

Both should be fixed together by wrapping the read→modify→write window with `fcntl.flock` on a sibling lock file — the locking helper covers state.ts and queue.ts uniformly. Tracked as the first item under "v1.5 priorities (correctness)" in #99 (the rolling notifications followups issue), with the fix paths spelled out.

Deferring this PR's queue race fix to v1.5 so the locking change is reviewed once across all writers, rather than landing partial coverage now.

src/notifications/format.ts
```diff
@@ -0,0 +1,28 @@
...
+const SEVERITY_PREFIX: Record<string, string> = {
+  info: "🐝",
+  warn: "⚠️",
+  error: "🚨",
+};
+
+function renderOne(n: Notification): string {
+  const prefix = SEVERITY_PREFIX[n.severity ?? "info"] ?? SEVERITY_PREFIX.info;
+  return `${prefix} ${n.title}\n${n.body}`;
+}
```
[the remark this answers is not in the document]
kaghni: Addressed in 76b707b — but with a docstring clarification rather than a content scrubber.

The rule's intent was that the renderer's OWN TEMPLATE shouldn't embed `"DEEPLAKE MEMORY"` / `"HIVEMIND"` (so we don't accidentally collide with the existing memory boilerplate block). It was NEVER about sanitizing notification content — admins are trusted, and stripping those substrings from arbitrary title/body text would mangle legitimate copy (e.g. a notification literally talking about "DEEPLAKE MEMORY"). 

Updated the docstring at `src/notifications/format.ts` to make this explicit: anti-pattern guard applies to the renderer's template, not to caller-supplied content. If we ever surface customer-controllable text, sanitation belongs at the source (admin endpoint), not at render time.

src/notifications/rules/welcome.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (!creds?.token) return null;
+    const userName = creds.userName ?? "there";
+    const orgName = creds.orgName ?? creds.orgId;
+    const workspace = creds.workspaceId ?? "default";
+    return {
+      id: "welcome",
+      severity: "info",
+      title: `Welcome back, ${userName}`,
+      body: `Connected to org ${orgName} (workspace ${workspace}).`,
+      dedupKey: { savedAt: creds.savedAt },
```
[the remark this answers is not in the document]
kaghni: Addressed in 76b707b — added the `?? "your org"` fallback at `src/notifications/rules/welcome.ts:22`:

```ts
const orgName = creds.orgName ?? creds.orgId ?? "your org";
```

Plus a regression test in 30bbe09 (`claude-code/tests/notifications-coverage.test.ts`) that constructs a malformed-credentials case (both `orgName` and `orgId` undefined) and asserts the rendered body contains `"your org"` and not `"undefined"`. Coverage on welcome.ts is now back to 100% across all metrics.

src/notifications/probe/probe-claude-code.js
```diff
@@ -0,0 +1,25 @@
...
+  process.exit(0);
+}
+
+// Drain stdin so Claude Code's writer doesn't EPIPE.
+let _ignored = "";
+process.stdin.on("data", (chunk) => { _ignored += chunk.toString(); });
+process.stdin.on("end", () => {
+  const stdoutToken = `PROBE_STDOUT_${randomUUID()}`;
+  const stderrToken = `PROBE_STDERR_${randomUUID()}`;
+  process.stderr.write(`${stderrToken}: if you see this on screen, stderr is visible at SessionStart\n`);
```
efenocchi: why do we have this?

src/notifications/rules/welcome.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import type { Rule } from "../types.js";
+
+export const welcomeRule: Rule = {
+  id: "welcome",
+  trigger: "session_start",
+  evaluate({ creds }) {
+    if (!creds?.token) return null;
+    const userName = creds.userName ?? "there";
```
efenocchi: ?

src/notifications/rules/welcome.ts
```diff
@@ -0,0 +1,28 @@
...
+    // Fallback chain: orgName → orgId → "your org". The "your org" arm
+    // guards against a malformed credentials.json where both fields are
+    // missing — without it, the rendered body would contain "undefined".
+    const orgName = creds.orgName ?? creds.orgId ?? "your org";
```
efenocchi: is it correct?

src/notifications/AGENT_CHANNELS.md
```diff
@@ -0,0 +1,83 @@
...
+
+| Agent | Adapter shipped? | Roadmap order |
+|---|---|---|
+| Claude Code | ✅ `delivery/claude-code.ts` (dual-channel: `systemMessage` + `additionalContext`) | shipped |
+| openclaw | ❌ — | next |
+| Codex | ❌ — | TBD |
+| Cursor | ❌ — | TBD |
+| Hermes | ❌ — | TBD |
+| Pi | ❌ — no SessionStart hook upstream | TBD |
+
```
efenocchi, line 19: should be here? why do we have this file?

src/notifications/probe/probe-claude-code.js
kaghni: Good catch — these were Phase-0 research scaffolding for sniffing how each agent's harness handles SessionStart hook stdout/stderr/JSON shapes. The findings (which channel each harness actually surfaces to the user vs. the model) are preserved in `src/notifications/AGENT_CHANNELS.md`; the probe scripts themselves are 50 lines each and trivially recreatable when we wire up the next agent.

Deleted in `10c4605`.

src/notifications/rules/welcome.ts
kaghni: Agreed — `"there"` reads awkwardly. Changed to drop the comma-clause entirely when `userName` is missing: title becomes `"Welcome back"` (no name) instead of `"Welcome back, there"`. Updated in `10c4605`.

src/notifications/rules/welcome.ts
kaghni: Right — exposing the `orgId` UUID in user-facing copy is bad UX (UUIDs are unreadable to humans, arguably worse than the original `undefined` leak). Removed `orgId` from the fallback chain entirely; it now reads `"your organization"` if `orgName` is missing. Fixed in `10c4605`.

src/notifications/AGENT_CHANNELS.md
kaghni: Fair question — it wasn't obvious from the file itself. The file is a record of empirical research that informed the v1 Claude Code adapter (and that the next agent integration — openclaw / codex / cursor / hermes / pi — will need). Each agent's harness handles SessionStart hook stdout/stderr/JSON shapes differently; rediscovering this from scratch costs an hour-plus per agent, so the findings live next to the code that depends on them.

I'd rather keep it (and pay the small file-tree cost) than rediscover the same things when we wire the next agent. To make the role obvious, I prepended an explanatory header in `10c4605`:

> **What this file is and why it lives here.** This is not runtime documentation — it's a record of the empirical research that informed the v1 Claude Code adapter and what each future agent integration will need. ... If the file ever drifts from reality, update it; if it stops being useful, delete it.

If you'd still rather move it out of `src/notifications/` (e.g. into a top-level `docs/` or `research/` dir), happy to relocate.

efenocchi (approved):

PR #97, opened 2026-05-06
feat(hooks): centralize autoupdate on the npm `hivemind update` command

kaghni:
## Summary

Centralizes hivemind autoupdate on the npm `hivemind update` command. Every agent's session-start now invokes a single shared helper (`src/hooks/shared/autoupdate.ts:autoUpdate`) which shells out to `hivemind update` — one mechanism, one source of truth, all agents.

Replaces three divergent legacy paths:
- **Claude Code**: `claude plugin update hivemind@hivemind --scope X` against marketplace + GitHub raw `package.json`
- **Codex**: `git clone --depth 1 --branch v<latest>` + `cp` against GitHub raw `package.json`
- **OpenClaw**: ClawHub registry probe + "openclaw plugins update" advice text

**Cursor / Hermes / pi** previously had no autoupdate at all; they pick it up for free here.

The trigger is per-agent (session-start), but the action is universal — `hivemind update` runs `npm install -g @deeplake/hivemind@latest` then re-execs `hivemind install --skip-auth`, refreshing **every detected agent on the machine** in one shot.

## OpenClaw exception

OpenClaw is the lone exception: its plugin bundle stubs out `node:child_process` for sandbox / static-analysis (`esbuild.config.mjs:279`), so it can't shell out. We centralize the OpenClaw plugin at:
- **Source-of-truth layer**: `registry.npmjs.org/@deeplake/hivemind/latest` instead of `clawhub.ai/api/v1/packages/hivemind`
- **Advice-text layer**: "Run: `hivemind update`" instead of "Run: `openclaw plugins update hivemind`"

Real upgrades happen when the user runs the suggested command, or when **any other agent's** session-start fires `autoUpdate` — that helper's `hivemind install --skip-auth` refresh covers OpenClaw too.

## Bonus bug fix: hook reorder

`autoUpdate` now fires **before** the per-hook DB ensure-table calls. Those calls can stall for tens of seconds against a slow / unreachable backend, and `autoUpdate` has zero dependency on table state. Without this reorder, a flaky network silently prevented the upgrade notice from ever appearing.

## Test plan

- [x] `npm test` — **1803 / 1803 pass**, no regressions
- [x] `npm run build` — bundles re-emitted, no drift
- [x] **End-to-end** verification: with `creds.autoupdate=false` and `HIVEMIND_DEBUG=1`, each agent's session-start hook produced these lines in `~/.deeplake/hook-debug.log`:
  ```
  [autoupdate] agent=<name> entered
  [autoupdate] agent=<name> skip: autoupdate=false
  ```
  Verified for: **claude, codex, cursor, hermes**. **pi** source-confirmed (inline implementation mirrors the shared helper; pi extension ships raw .ts and can't import shared modules). **OpenClaw** bundle grep verified npm registry URL is in place.
- [x] **Negative-pattern checks** in tests verify no occurrences of `claude plugin update`, `git clone`, `openclaw plugins update`, or `clawhub.ai` in helper output

## Files

**New:**
- `src/hooks/shared/autoupdate.ts` — the shared helper

**Modified (replaced legacy autoupdate, added autoUpdate call):**
- `src/hooks/session-start.ts` — Claude Code
- `src/hooks/codex/session-start-setup.ts` — Codex
- `src/hooks/cursor/session-start.ts` — Cursor (first time has any autoupdate)
- `src/hooks/hermes/session-start.ts` — Hermes (first time)
- `pi/extension-source/hivemind.ts` — pi (inline, since pi extension ships raw .ts)
- `openclaw/src/index.ts` — npm registry + `hivemind update` advice; notice-only, no spawn

**Tests:**
- `claude-code/tests/autoupdate.test.ts` — 26 tests for the helper: gating, spawn shape, output parsing, per-agent restart hints, negative patterns
- `claude-code/tests/{session-start,codex-session-start-setup,cursor-session-start,hermes-session-start}-hook.test.ts` — updated to mock `autoUpdate` at the boundary, removed legacy assertions, added negative-pattern guards

## Plan reference

Plan: `/home/ubuntu/.claude/plans/1-denied-we-do-bright-elephant.md` — per-agent audit table + design rationale.

## Release Notes

* **New Features**
  * Automated update system for Hivemind binary during session startup with clear restart guidance per agent.
  * Enhanced update notices displaying current installed version when updates are available.
  * Improved setup logging and credential handling with better error visibility.

* **Refactor**
  * Centralized update mechanism across all agents with improved reliability and consistent behavior.


openclaw/src/index.ts
```diff
@@ -170,13 +168,16 @@ async function checkForUpdate(logger: PluginLogger): Promise<void> {
   try {
     const current = getInstalledVersion();
     if (!current) return;
-    const res = await fetch(VERSION_URL, { signal: AbortSignal.timeout(3000) });
+    const res = await fetch(VERSION_URL, { signal: AbortSignal.timeout(5000) });
     if (!res.ok) return;
     const latest = extractLatestVersion(await res.json());
     if (latest && isNewer(latest, current)) {
-      logger.info?.(`⬆️ Hivemind update available: ${current} → ${latest}. Run: openclaw plugins update hivemind`);
+      pendingUpdate = { current, latest };
```
efenocchi, line 196: does it always work? is it deterministic?

src/hooks/shared/autoupdate.ts
efenocchi, line 1: tbh I'm not sure about the cache, in the worst scenario:
- the check runs (i.e. time 13:00)
- we update the version (time 13:01)  
- we'll be able to check if it worked in 4h (time 17:00)
(if I'm not mistaken)

It would be bad for the users and for internal testing

efenocchi: It might make sense to use an internal variable so it's always 0 for new sessions.

claude-code/tests/session-start-hook.test.ts
```diff
@@ -224,91 +202,62 @@ describe("session-start hook — placeholder branching", () => {
...
+    // node:child_process isn't mocked anymore.
     await runHook();
-    expect(debugLogMock).toHaveBeenCalledWith(
-      expect.stringContaining("autoupdate snapshot outcome:"),
-    );
+    // No assertion needed — if execSync were re-introduced and called,
+    // it'd hit the real binary and likely fail or hang. The presence of
+    // autoUpdateMock being called proves the new path is in use.
+    expect(autoUpdateMock).toHaveBeenCalled();
   });
```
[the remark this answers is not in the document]
kaghni: Addressed in commit f0ffd9e — dropped the misleading execSync test entirely. The replacement `fetch` negative-pattern test still proves the legacy code path is gone (the legacy autoupdate flow ALWAYS started with `fetch(githubraw)`; if that doesn't fire, the execSync that was gated on its result can't fire either). Comment in the test file documents the vitest ESM `vi.spyOn` limitation that prevented the cleaner direct-spy approach.

openclaw/src/index.ts
```diff
@@ -765,22 +763,31 @@ export default definePluginEntry({
...
           const current = getInstalledVersion();
           if (!current) return;
-          const res = await fetch(VERSION_URL, { signal: AbortSignal.timeout(3000) });
+          const res = await fetch(VERSION_URL, { signal: AbortSignal.timeout(5000) });
           if (!res.ok) return;
           const latest = extractLatestVersion(await res.json());
           if (!latest || !isNewer(latest, current)) return;
           pendingUpdate = { current, latest };
-          logger.info?.(`Hivemind update available: ${current} → ${latest}. Agent will be prompted to install when user asks.`);
+          logger.info?.(`⬆️ Hivemind update available: ${current} → ${latest}. Run: hivemind update`);
```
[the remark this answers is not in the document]
kaghni: Addressed in commit b5d2972 — collapsed to a single path. Removed the inline async block; `checkForUpdate(logger)` now owns `pendingUpdate` and respects `config.autoUpdate`. The single call from `register()` is gated behind `if (config.autoUpdate !== false)`, so disabled installs don't probe and enabled installs probe exactly once.

pi/extension-source/hivemind.ts
```diff
@@ -644,6 +644,32 @@ export default function hivemindExtension(pi: ExtensionAPI): void {
...
+            child.stderr?.on("data", d => { out += d.toString(); });
+            child.on("close", () => {
+              const m = out.match(/Updated to .+\./);
+              if (m) process.stderr.write(`✅ Hivemind ${m[0]} Restart pi to apply.\n`);
+              resolve();
+            });
+            child.on("error", () => resolve());
+          });
+        }
+      } catch { /* network down / which missing — silent */ }
```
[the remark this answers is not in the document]
kaghni: Addressed in commit b5d2972 (with the exact diff from the suggestion). Added `autoupdate?: boolean` to pi's `Creds` interface and threaded `parsed.autoupdate` through `loadCreds()`. `hivemind autoupdate off` now correctly suppresses the inline updater on pi.

src/hooks/shared/autoupdate.ts
```diff
@@ -0,0 +1,160 @@
...
+  if (!summary) return;
+
+  // Surface upgrade outcomes; suppress "is up to date" (common case,
+  // would spam stderr on every session-start).
+  if (/Updated to/.test(summary)) {
+    stderr(`✅ Hivemind ${summary} ${RESTART_HINT[opts.agent]}\n`);
+  } else if (/Update available/.test(summary)) {
+    stderr(`⬆️ Hivemind: ${summary}\n`);
+  }
+}
```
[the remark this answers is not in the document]
kaghni: Captured as a follow-up. Initial draft of this PR did add a flock-style lock at `~/.deeplake/.autoupdate.lock` inside this helper. After the latency-fix rewrite to a fire-and-forget detached spawn (commit 601aa5c), the helper releases the lock immediately after dispatching the child — defeating the purpose, since the actual upgrade is the long-running detached process. The proper home for the lock is `src/cli/update.ts:runUpdate()` so the long-running `hivemind update` process owns the lock for its full lifetime. Tracked separately rather than blocking this PR; `npm install -g` has its own internal locking that prevents corruption in the meantime, so concurrent invocations either both no-op (already at latest) or one wins idempotently.

openclaw/src/index.ts
kaghni: Tracked as issue #105 — during real-world E2E we hit `AbortSignal.timeout(5000)` once on cold gateway init. Steady-state registry latency measured 167ms (`time curl https://registry.npmjs.org/@deeplake/hivemind/latest`), so 5s is plenty after warm-up — the cold-init failure is the issue. Fix is a small bump to 10s + optional retry-on-AbortError. Out of scope for this PR (it's a pre-existing concern surfaced by E2E, not something this PR caused or worsens).

src/hooks/shared/autoupdate.ts
kaghni: Good catch — you're right, that 4h gap would bite both end users on a fresh release and us during internal testing. Removed the cache entirely in commit 6a50520. Reasoning: the cache only saved background CPU (an npm GET inside the spawned process), it never affected session-start latency since the dispatch is already detached. So it wasn't earning its keep on the hot path. Now every session-start fires the detached `hivemind update`; the spawned process checks npm itself and exits silently if up-to-date. Trade-off: 1 npm registry GET per session-start instead of 1 per 4h, in exchange for ~zero "miss new release" window. Concurrency safety is handled by `npm install -g`'s internal locking + the follow-up flock in `runUpdate()` (CodeRabbit thread above).

kaghni: Made obsolete by removing the cache entirely in commit 6a50520 (see thread above). For context on why an in-memory variable wouldn't have helped even if we kept the cache: each session-start hook is its own short-lived Node process (Claude Code spawns `session-start.js` and `session-start-setup.js` as separate child processes), so there's no shared in-memory state between hook invocations to cache against. Each invocation either always checks (current behavior, no cache) or persists state to disk (the 4h-cache approach we just dropped).

efenocchi (approved):

PR #98, opened 2026-05-06
feat(skilify): background skill-mining worker + Deeplake skills table + per-agent gate

efenocchi:
## Summary

Adds **skilify** — a background worker that mines recent agent sessions in Deeplake, gates whether each cluster is worth crystallising as a reusable skill, and writes `SKILL.md` files (locally + to a new `skills` Deeplake table for org-wide provenance). Plus a `hivemind skilify pull` command so teammates can fetch each other's skills without git roundtrips.

Works for **all 4 agents** (Claude Code / Codex / Cursor / Hermes), each using its own headless CLI for the gate call — no hard `claude` dependency for non-Claude users.

## Architecture

```
Stop hook (or SessionEnd)
   ↓ counter ≥ 20 (or always-fire on SessionEnd)
spawn-skilify-worker (detached subprocess)
   ↓
skilify-worker.ts
   ├─ fetch last 10 sessions in scope from Deeplake `sessions` table
   ├─ extract Q&A pairs (drop tool_call / thinking)
   ├─ gate: claude / codex / cursor-agent / hermes -p (per-agent CLI)
   │   → JSON verdict: KEEP <name> <body> | SKIP <reason> | MERGE <name> <body>
   ├─ KEEP   → writeNewSkill   (<project>/.claude/skills/<name>/SKILL.md)
   ├─ MERGE  → mergeSkill       (bump version, dedup source_sessions)
   └─ insertSkillRow            (append-only row in `skills` table)
```

## What's in the box

- **Per-project state** (`~/.deeplake/state/skilify/<projectKey>.json`): counter, watermark, generated skill names. Atomic RMW lock. Project key = sha1 of git remote (with cwd-hash fallback).
- **Agent-agnostic extractor**: all 4 agents normalise to the same `{user_message, assistant_message, tool_call}` taxonomy in Deeplake — one extractor handles all.
- **Per-agent gate dispatch**: `gate-runner.ts` invokes `claude`, `codex`, `cursor-agent`, or `hermes` based on the originating agent. Hermes via OpenRouter requires `OPENROUTER_API_KEY` in env (inherited from parent).
- **Append-only `skills` table** on Deeplake: `(project_key, name)` keyed, `version` bumps on each MERGE. `ORDER BY version DESC LIMIT 1` reads the current state. Sidesteps the UPDATE-coalescing quirk (CLAUDE.md note).
- **Lazy table create**: first INSERT against a workspace where the table doesn't exist runs `CREATE TABLE` inline and retries. No migration step.
- **Hook integration** for all 4 agents via shared `triggers.ts`:
  - `tryStopCounterTrigger` (mid-session, threshold 20 by default)
  - `forceSessionEndTrigger` (always-fire backup at session end — also handles `claude -p` mode where Stop is async-racey)
- **`hivemind skilify` CLI**:
  - `scope <me|team|org>`, `team add/remove/list`
  - `install <project|global>`, `promote <skill-name>`
  - **`pull [skill-name] [--to project|global] [--user X | --users a,b,c | --all-users] [--dry-run] [--force]`** — fetch skills from Deeplake to local FS, with conflict handling (backup `.bak`, version compare).
- **`/skilify` slash command** for Claude Code + Codex (forwards `$ARGUMENTS` to the CLI).

## Files

- New: `src/skilify/{state,extractors/index,scope-config,skill-writer,gate-parser,gate-runner,skills-table,spawn-skilify-worker,skilify-worker,triggers,pull}.ts` (~1600 LOC)
- New: `src/commands/skilify.ts` (CLI dispatcher, ~280 LOC)
- New: `claude-code/commands/skilify.md` + `codex/commands/skilify.md` (slash command wrappers)
- Modified: `src/config.ts` (skillsTableName), `src/deeplake-api.ts` (ensureSkillsTable), `src/cli/index.ts` (subcommand dispatch), 7 hook files for per-agent integration, `esbuild.config.mjs` (4 bundle outputs), `README.md` (Skills section).

## Testing

- **10 new unit-test files**, ~200 new tests, all passing alongside the existing suite (**1911/1911 total**, 102 files).
- **Coverage** for skilify modules: every file ≥87% lines (most ≥95%); branch coverage 60-100% with the lower numbers explained in `vitest.config.ts` (subprocess error paths + env fallbacks). Per-file thresholds added so PRs touching these files keep the bar.
- **End-to-end matrix** (not in vitest — too heavy): script at `/tmp/skilify-e2e-matrix.mjs` runs 7 cases against the `plugin_test_1/test1` sandbox:
  - KEEP × 4 agents (claude_code / codex / cursor / hermes — each invoking its own CLI)
  - install=global path
  - SKIP path (gate sees existing skill, skips)
  - MERGE path (gate extends existing skill to v2)

  Last full run: **7/7 passed**. Each `source_agent` column in the resulting Deeplake row matches the originating agent.

## Test plan

- [ ] CI: build + 1911 tests pass
- [ ] Coverage report: skilify modules ≥ thresholds (lines ≥80, most ≥90)
- [ ] Smoke: `node bundle/cli.js skilify --help` shows full usage
- [ ] Smoke: `node bundle/cli.js skilify pull --dry-run` against sandbox returns row count + filter
- [ ] Manual: `claude --plugin-dir claude-code` interactive session, send 4-5 prompts on a coherent theme, `/exit`, verify `~/.claude/hooks/skilify.log` shows `verdict=KEEP|SKIP|MERGE` and a SKILL.md is written under the project's `.claude/skills/`.

## Deferred (not v1)

- `/skilify --consolidate` — semantic-similarity pass over all project skills to catch merges the per-trace gate missed.
- Auto-pull on SessionStart (with notification "N new skills available").
- Cross-project skill clustering for org-wide skills.
- Per-agent install location for global skills (today `install=global` always writes to `~/.claude/skills/`; ideally cursor/hermes users land in their respective globals).

## Known upstream caveat

`claude -p` one-shot mode doesn't capture `assistant_message` rows reliably (Stop hook is async, parent exits before the INSERT lands). Hits the wiki worker too — same root cause. Skilify works fine in interactive mode and in the scenarios above; document'd in the README. Separate issue worth filing.

## Release Notes

* **New Features**
  * Added Skilify—a skill mining system that automatically generates reusable Claude Code skills from recent sessions and writes them as SKILL.md files.
  * Added `skilify` CLI command with subcommands for managing scope, team membership, skill installation, versioning, and remote skill pulling.
  * Introduced per-project and global skill management with version control and source session tracking.

* **Documentation**
  * Added Skilify documentation to README and command help.


codex/commands/skilify.md
```diff
@@ -0,0 +1,13 @@
+---
+description: View or change the skilify scope (me / team / org) and team list
+allowed-tools: Bash
+argument-hint: "[scope <me|team|org> | team add <name> | team remove <name> | team list | status]"
+---
+
+Run the skilify scope/team management command and report the output verbatim:
+
+```bash
+hivemind skilify $ARGUMENTS
```
kaghni: Why do we have skillify.md for each agent? Is it going to work only with hivemind cli or it is also going to work inside agents with /hivemind command?

efenocchi:
/claude-review

Triggered after fully validating the skilify e2e chain end-to-end against the live Deeplake sandbox (`plugin_test_1/test1`) — 2027/2027 unit + 10/10 Pi real-runtime + 6/6 OpenClaw bundle + 65/65 pull cherry-pick + 6/6 Pi worker bundle. Branch HEAD: `5d6dfb6`. Two recent fixes worth re-reviewing: gate-runner missing `pi` case (silently SKIPped every cluster on Pi) and a redundant Pi spawn-side lock (limited mining to 1 spawn per runtime lifetime). Both flagged here so the reviewer doesn't have to re-derive the chain.

codex/bundle/stop.js
```diff
@@ -1084,6 +1379,13 @@ async function main() {
     }
     throw e;
   }
+  forceSessionEndTrigger({
+    config,
+    cwd: input.cwd ?? "",
+    bundleDir: bundleDirFromImportMeta(import.meta.url),
+    agent: "codex",
+    sessionId
+  });
```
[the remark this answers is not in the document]
efenocchi: Acknowledged — keeping this as a deliberate design choice rather than reverting. Rationale:

**Codex's Stop hook fires on every assistant turn, not on session end.** Codex doesn't have a separate `session_end` event the way Claude Code does. The `Stop` event is the only signal we get that "the agent finished a turn", and the "session ended" semantics is inferred from the user closing the codex process — which doesn't fire anything we can hook.

If we treated Codex `Stop` as a per-turn counter (the `tryStopCounterTrigger` path used for CC's mid-session trigger), we'd never fire skilify mining at the end of a Codex session. The user would close Codex, no event fires, the worker never runs. The session's prompt+answer pairs sit in the sessions table forever, never minable except by chance through another agent's worker that happens to share the project.

`forceSessionEndTrigger` is the right primitive here because it's explicitly the "fire once, regardless of counter" path. Calling it on every Codex Stop is "over-triggering" only in the sense that it fires more often than strictly necessary — but the per-projectKey worker lock at `~/.deeplake/state/skilify/<projectKey>.lock` (with 10-min stale fallback in `state.ts:tryAcquireWorkerLock`) makes the over-triggering cheap:

  - First Codex Stop in a session → worker spawns, mines, advances watermark, releases lock
  - Subsequent Stops in the same session → worker spawns, takes the lock, finds nothing past watermark, exits in <1s, releases the lock
  - Cumulative cost per long Codex session: ~1-2s of redundant SQL queries spread across the session, no correctness impact

This is the same trade-off OpenClaw makes (filed as #100), and matches the pattern across Cursor / Hermes / Pi which all fire mining on their respective per-turn events. The only agent with a "real" session_end is Claude Code, and even there we use both paths (counter for periodic + force on session end) to handle the case where the counter never crossed the threshold.

The in-code comment at `src/hooks/codex/stop.ts:170-180` already explains the rationale to future readers; happy to expand it if you think the explanation needs to be more prominent. Not changing the call shape itself though — `tryStopCounterTrigger` would silently break Codex mining on every short session.

— Posted as part of PR #98 final review pass. CI green at HEAD `e823063`.

openclaw/skills/SKILL.md
```diff
@@ -45,6 +45,25 @@ Do NOT jump straight to reading raw JSONL files. Always start with `hivemind_ind
 - `/hivemind_update` — shows how to install (ask the agent, or run `openclaw plugins update hivemind` in your terminal)
 - `/hivemind_autoupdate [on|off]` — toggle the agent-facing update nudge (on by default: when a newer version is available, the agent is prompted to install it via `exec` if you ask to update)
 
+## Skill Management (skilify)
+
+Hivemind also mines reusable Claude skills from agent sessions and stores them in a per-org Deeplake table. Openclaw itself doesn't run sessions to mine, but you can pull skills others have already mined for the user. These run in the user's terminal (the openclaw plugin does not register them as `/hivemind_*` commands):
+
+- `hivemind skilify` — show scope/team/install + per-project state
```
kaghni, line 52: for openclaw the commands have "/" syntax to work on telegram (or whatsapp) did you test in openclaw were those commands working without "/"?

PR #112, opened 2026-05-08
Auto-pull all-author skills at SessionStart

efenocchi:
## Summary

Every supported agent (Claude Code, Codex, Cursor, Hermes, pi) now auto-runs the equivalent of `hivemind skilify pull --all-users --to global` at the start of each session, so teammate-mined skills become available without anyone having to remember to run pull manually.

## Design decisions

- **Cadence**: throttled at the SessionStart entry point. The previous-run timestamp is persisted at `~/.deeplake/state/skilify/autopull-last-run.json` via an atomic tmp+rename write (mirrors `src/skilify/state.ts`). Default window is 30 minutes; configurable via `HIVEMIND_AUTOPULL_INTERVAL_MIN` (`0` forces every session, `-1` disables). `HIVEMIND_AUTOPULL_DISABLED=1` is the hard opt-out.
- **Failure mode**: never block SessionStart. `maybeAutoPull` catches every network / missing-table / auth error and resolves with `skipped: true`. The hook callers do not wrap it in try/catch — the function is documented as never-rejecting, which keeps the hooks' branch-coverage numbers intact.
- **Bound**: the pull runs synchronously (awaited) but is bounded by a 5-second timeout via `Promise.race`. A slow Deeplake never freezes the session.
- **Scope**: `install: "global"`, `users: []`, `force: false`. Global so all projects benefit. The last-run timestamp is only written on a successful pull, so a failed attempt is naturally retried on the next SessionStart.
- **First-run / not-logged-in**: silent skip (no nag).

## What got wired

New module `src/skilify/auto-pull.ts` encapsulates throttle + `loadConfig` + `runPull` + timestamp persistence. It accepts injected `loadConfigFn`, `queryFn`, `nowMs`, `timeoutMs`, `install`, `cwd` so tests are fully hermetic at the network seam.

`maybeAutoPull()` is called from each of the five SessionStart hooks right after credential bootstrap and before `additionalContext` is assembled:

- `src/hooks/session-start.ts` (Claude Code)
- `src/hooks/codex/session-start.ts` (Codex)
- `src/hooks/cursor/session-start.ts` (Cursor)
- `src/hooks/hermes/session-start.ts` (Hermes)
- `pi/extension-source/hivemind.ts` (pi — logic is **inlined** because pi ships as raw `.ts` with zero non-builtin runtime deps; mirrors the existing `spawnPiSkilifyWorker` inlining pattern)

## Tests

`claude-code/tests/skilify-auto-pull.test.ts` (21 cases) covers:

- `readIntervalMs`: defaults, integer minutes, `0`, negative, garbage, empty
- `writeLastRun` + `readLastRun` round-trip incl. malformed JSON paths
- `HIVEMIND_AUTOPULL_DISABLED=1` short-circuits before `loadConfig`
- `HIVEMIND_AUTOPULL_INTERVAL_MIN=-1` short-circuits before `loadConfig`
- not-logged-in silent skip (`loadConfig` returns null)
- throttle: first call writes timestamp, second call inside window is skipped (no SQL, no timestamp bump), call past the window runs again, `interval=0` forces every call
- failure paths: query throw → swallowed, timestamp NOT updated; `relation "skills" does not exist` → treated as empty, timestamp updated; hanging query is bounded by an injected 100ms timeout and completes well under 1s
- `install=global` default writes under `~/.claude/skills`

All assertions go through the real `maybeAutoPull` → `runPull` pipeline with a mocked `QueryFn`, per `CLAUDE.md`'s "mock at the network seam" rule. Filesystem effects are scoped to a per-test temp `HOME` so the developer's real `~/.deeplake` is untouched.

`claude-code/tests/codex-session-start-hook.test.ts` adds a stub for `src/skilify/auto-pull.js` so the codex hook test (which doesn't mock `src/config.js` or `src/deeplake-api.js`) doesn't fire a real network query during SessionStart. The other agents' tests already mock `loadConfig` + `DeeplakeApi`, so their auto-pull invocation is naturally contained.

## Coverage

`src/skilify/auto-pull.ts`: **98% lines / 100% functions / 87% branches**. `vitest.config.ts` adds `80/70/90/80` thresholds for the new file (matches the rest of the skilify suite).

Full suite: **2099 / 2099 passing**, no threshold ERRORs.

## Docs

- README "Skills (skilify)" section gains an "Auto-pull at SessionStart" sub-section with the env-var table.
- `hivemind skilify pull` USAGE block in `src/cli/index.ts` mentions the auto-pull behaviour and the opt-out env vars.

## Test plan

- [ ] Land this PR before #111 merges, OR rebase if #111 lands first (this branch only adds a new file under `src/skilify/` and does not touch the files PR #111 edits)
- [ ] Verify SessionStart timing on a real Claude Code session — should add at most 5 seconds on the slowest cold path, normally sub-100ms once throttled
- [ ] Confirm the timestamp file appears at `~/.deeplake/state/skilify/autopull-last-run.json` after the first session
- [ ] Confirm a second SessionStart within 30 min skips the pull (debug log line `autopull: throttled (...)` in `~/.deeplake/hook-debug.log` when `HIVEMIND_DEBUG=1`)
- [ ] Smoke-test `HIVEMIND_AUTOPULL_INTERVAL_MIN=0` and `HIVEMIND_AUTOPULL_DISABLED=1` opt-outs
- [ ] Smoke-test against pi (extension is shipped as raw `.ts`; verify `autopullSkillsForPi` runs at session_start without breaking the existing capture path)

## Summary by CodeRabbit

* **New Features**
  * Automatic skills auto-pull at session start: Agents now automatically sync org-wide reusable skills on each session initialization with configurable throttling and timeout behavior. Control via `HIVEMIND_AUTOPULL_INTERVAL_MIN` and `HIVEMIND_AUTOPULL_DISABLED` environment variables.

* **Documentation**
  * Updated documentation for auto-pull configuration and environment variables.

* **Bug Fixes**
  * Improved schema migration robustness and reduced excessive database operations.


efenocchi:
## Re: coderabbit suggestion to fire-and-forget `maybeAutoPull()` in session-start hooks

We're keeping the synchronous `await maybeAutoPull()` in codex / cursor / hermes deliberately. Two reasons:

1. **The user-visible UX is "skills available before the first prompt".** If we make the pull fire-and-forget, the user's first prompt can land on the model before pulled skills are written to disk, so the freshly-mined teammate skill they just opened the session to use isn't discoverable until next session. Bad cache hit rate on the very thing this PR exists to enable.

2. **The wait is bounded.** `maybeAutoPull` enforces a 5-second internal `Promise.race` timeout. A slow Deeplake never freezes session-start past that. The 30-min throttle means ~1 pull per session-start in the worst case; once throttled, the call returns synchronously without I/O. Net cost per cold session: ≤5s, almost always sub-second when the table is small.

The recently-landed pi commit (9914adf) follows the same pattern via `spawnSync` with a 6s wall-clock cap (5s inner timeout + 1s spawn overhead) — same semantics, different transport because pi's extension ships as raw .ts and can't import the shared module.

Closing this finding as "won't fix" — the suggestion would change documented behaviour, not a bug. Coderabbit didn't read the PR description's "Bound" / "Failure mode" sections.

src/skilify/auto-pull.ts
```diff
@@ -0,0 +1,127 @@
...
+  });
+}
+
+/**
+ * Top-level entry. Decides whether to skip (env / not-logged-in) and otherwise
+ * runs a bounded, all-failures-swallowed pull.
+ *
+ * Always resolves; never rejects. The return value is informational only.
+ */
+export async function maybeAutoPull(deps: AutoPullDeps = {}): Promise<AutoPullResult> {
```
kaghni: cc also suggested me name with maybe, but I would rename it to just autoPull. being able to disable it doesn't make it maybe

kaghni (approved):

Issue #118, opened 2026-05-11
skills table: add `contributors` column for team/org scope edits


efenocchi:
## Context

The `skills` Deeplake table (schema in `src/skilify/skills-table.ts`) stores a single `author` column. When the gate decides `MERGE`, a new row is inserted with `author = cfg.userName` of whoever ran the mining — overwriting the lineage of who has actually contributed to the skill over time.

Today the `scope` column on each row is effectively cosmetic — `org` is unused in practice (`scope=org` never observed in prod data; nobody on the team has set it via `hivemind skilify scope org`), and `authorClause()` in `skilify-worker.ts` is untested at the worker level. The model below uses only `me` and `team`; `org` is dropped from the data model (CLI may keep it as a no-op back-compat alias for `team`).

## Proposal

Two coupled changes:

### A. Add `contributors` column to the `skills` table

JSON array of usernames, ordered chronologically by edit.

```sql
ALTER TABLE skills ADD COLUMN IF NOT EXISTS contributors TEXT NOT NULL DEFAULT '[]';
```

- **Schema**: extend `createSkillsTableSql()` so fresh tables include the column. Existing tables get it via `ADD COLUMN IF NOT EXISTS` on first insert (back-compat — old rows simply read as `[]`).
- **Read**: `JSON.parse(row.contributors)` with fallback to `[row.author]` for legacy rows (empty array → treat author as the single contributor).
- **Write (KEEP)**: initialize to `[cfg.userName]`.
- **Write (MERGE)**: read the previous version's `contributors`, append `cfg.userName` if not already present, write the new array on the v+1 row.
- **Display**: `hivemind skilify` shows `author (+N contributors)` when the array has more than one entry. Pulled SKILL.md frontmatter can mirror the array as a `contributors:` block for at-a-glance authorship.

### B. Auto-promote `scope` on cross-author MERGE

`scope` becomes an *observed* fact (derived from edit pattern), not a user-set preference:

- KEEP by `cfg.userName` on a fresh skill → `scope = "me"`, `contributors = [cfg.userName]`.
- MERGE where the matched row's `author == cfg.userName` → `scope` unchanged, `contributors` += self if not present.
- MERGE where the matched row's `author != cfg.userName` → `scope = "team"`, `contributors = [...previous, cfg.userName]` (cross-author edit is the signal that the skill is becoming shared).
- No further auto-promotion above `team` — `org` is dropped from the model.

[...]

PR #119, opened 2026-05-11
fix(skilify): plug the three holes that produced cross-author duplicate skills

efenocchi:
## Summary

Three tactical fixes for the cross-author and cross-project duplicate skills problem on the `skills` Deeplake table. Each commit is independent and focused; landing all three closes the gap that produced the duplicates we currently have in prod (e.g. `meta-harness-continual-learning` ↔ `continual-learning-meta-harness`, the three near-identical `hivemind-*-testing` skills).

### What was actually broken

1. **`created_at` was reset on every MERGE.** All 27 rows in the prod `skills` table have `created_at == updated_at` because the worker stamped `now()` on both fields at every INSERT. The local SKILL.md frontmatter already preserved the v=1 creation date correctly — only the DB row didn't see it.

2. **`deriveProjectKey` hashed the raw git remote URL.** Five surface forms of the same repo (`git@...`, `https://...`, with/without `.git`, with embedded credentials) produced five different `project_key` values. Two devs cloning the same repo with different URL styles ended up with disjoint state.

3. **The gate only read `<cwd>/.claude/skills/`, never `~/.claude/skills/`.** Autopull lands skills under the global root regardless of scope-config; the worker only looked at the project root. With the default `install=project`, the gate saw zero existing skills no matter how many had been autopulled — so every mining run started from a blank slate and re-created near-identical skills.

## Commits

- `fix(skilify): preserve created_at across MERGE in skills table` — thread the v=1 `created_at` through `SkillWriteResult` instead of stamping `now()`.
- `fix(skilify): normalize git remote URL before hashing for projectKey` — collapse SSH/HTTPS/.git/cred variants to a canonical `host/owner/repo` form.
- `fix(skilify): gate reads project + global skills, only [project] mergeable` — extract the existing-skills logic into a testable module, read both roots, tag entries `[project]` (MERGE-eligible) or `[global, read-only]` (reference only). Cross-author MERGE intentionally stays forbidden in this PR.

## Out of scope (filed separately)

- Cross-author MERGE with a `contributors` column and auto-promote `me → team` on cross-author edits — see activeloopai/hivemind#118. The interim restriction in this PR (MERGE only into your own project skills) avoids silently overwriting teammates' work until that lands.
- Compaction of old version rows (e.g. 7 `skills-autopull-fanout-symlinks` versions) — purely a storage-bloat concern, not a correctness one.

## Test plan

- [x] `npm run build` clean
- [x] `npm test` — 2189 passing (was 2175 before; +14 from new tests)
- [x] New tests: `skilify-state.test.ts` (URL normalization, 4 cases), `skilify-skill-writer.test.ts` (createdAt threading through `SkillWriteResult`), `skilify-existing-skills.test.ts` (7 cases: both roots, dedupe project-wins, source tagging, char-cap placeholder)
- [x] Bundle-scan tests updated for new gate-prompt heading
- [ ] Manual smoke: run a session in a project with autopulled global skills, confirm the gate prompt now mentions them as `[global, read-only]`

## Release Notes

* **New Features**
  * Skills now include creation and modification timestamps for better tracking.
  * Enhanced existing skills display showing both project and global read-only skills with clear source labels.

* **Improvements**
  * Clarified merge eligibility rules in skill curation prompts—only project skills are valid merge targets; global skills are reference-only references.
  * Improved project identification through consistent normalization of Git remote URLs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/119)


claude-code/bundle/session-end.js
```diff
@@ -353,15 +353,30 @@ function statePath(projectKey) {
...
+  if (!hadScheme) {
+    const scp = s.match(/^(?:[^@/\s]+@)?([^:/\s]+):(.+)$/);
+    if (scp)
+      s = `${scp[1]}/${scp[2]}`;
+  }
+  s = s.replace(/^[^@/]+@/, "");
+  s = s.replace(/\.git\/?$/i, "");
+  s = s.replace(/\/+$/, "");
+  return s.toLowerCase();
+}
```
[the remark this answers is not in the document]
efenocchi: Fixed in `886e6b5` — `normalizeGitRemoteUrl` now strips the default port for the scheme (`:443` for https, `:22` for ssh, `:9418` for git, `:80` for http). Non-default ports stay (an internal `:8443` is load-bearing, can't be dropped). Implemented in pure regex rather than `new URL()` because Node's `URL` class only auto-normalizes default ports for the "special" schemes (http/https/ws/wss/ftp/file) — git:// and ssh:// would still need manual handling, so a single regex pass is uniform.

Test cases for `:443`, `:22`, `:9418`, `:80` (collapsed) and `:8443`, `:2222` (preserved) added in the same commit.

src/skillify/state.ts
```diff
@@ -54,14 +54,51 @@ function lockPath(projectKey: string): string {
...
   const project = basename(cwd) || "unknown";
   let signature: string | null = null;
   try {
-    signature = execSync("git config --get remote.origin.url", {
+    const raw = execSync("git config --get remote.origin.url", {
       cwd, encoding: "utf-8", stdio: ["ignore", "pipe", "ignore"],
-    }).trim() || null;
+    }).trim();
+    signature = raw ? normalizeGitRemoteUrl(raw) : null;
   } catch {
```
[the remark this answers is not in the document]
efenocchi: Fixed in `886e6b5` — `normalizeGitRemoteUrl` now strips the default port for the scheme (`:443` for https, `:22` for ssh, `:9418` for git, `:80` for http). Non-default ports stay (an internal `:8443` is load-bearing, can't be dropped). Implemented in pure regex rather than `new URL()` because Node's `URL` class only auto-normalizes default ports for the "special" schemes (http/https/ws/wss/ftp/file) — git:// and ssh:// would still need manual handling, so a single regex pass is uniform.

Test cases for `:443`, `:22`, `:9418`, `:80` (collapsed) and `:8443`, `:2222` (preserved) added in the same commit.

src/skillify/state.ts
```diff
@@ -55,14 +55,72 @@ function lockPath(projectKey: string): string {
   return join(STATE_DIR, `${projectKey}.lock`);
 }
 
+/**
```
kaghni, line 58: why do we have this?

kaghni: where do we use git for skills and why

kaghni:
openclaw is missing

src/skillify/state.ts
efenocchi: `deriveProjectKey` produces a stable per-project identifier used in three places:

1. **`skills` table partition key** — every row carries `project_key`. Combined with `name` it's the dedup key the gate uses to decide KEEP vs MERGE, and what `pull.ts` uses to namespace skills cross-project (`<root>/<project_key>/...` in the legacy layout, now flattened to `<name>--<author>/`).
2. **Per-project state file** — `~/.deeplake/state/skillify/<project_key>.json` (Stop counter, watermark UUID, last mined date). Two projects mustn't share this file.
3. **Autopull manifest** — `pulled.json` records each pulled skill with its `project_key` so `unpull` knows which entries belong to which project.

**Why git remote URL specifically as the input to the hash:** it's the most reliable cross-machine identifier of "the same logical project". A repo cloned by Alice into `/Users/alice/work/foo` and by me into `/home/emanuele/code/foo` shares `git remote.origin.url` — so both their skills get the same `project_key`, the dedup gate can reason about them as one project, and pulling Alice's skill lands it in my namespace correctly.

If we hashed the absolute `cwd` instead, every clone of the same repo on every machine would have a different key → duplicate skills across cloners, no cross-cloner dedup. The cwd is only a fallback when there's no git remote (e.g. mining in a non-repo working dir like `/home/ec2-user`) — and that fallback is exactly what produced some of the duplicates we cleaned up in this PR (the three `hivemind-*-testing` skills emanuele had: one project cwd had `git remote`, one didn't, so they hashed differently).

The fix in this PR (`normalizeGitRemoteUrl`) makes the git-URL path actually deliver on that promise — before the fix, SSH vs HTTPS vs `.git`-suffix variants of the same URL were producing 5 different hashes for the same repo (test cases in `claude-code/tests/skillify-state.test.ts`).

efenocchi:
Good catch — addressed in `529a26f` (pushed right before your comment landed). `openclaw` is now in the `AGENTS` const of `skillify-bundle-scan.test.ts`, with a small special-case in `bundlePath()` to point at `openclaw/dist/` instead of `openclaw/bundle/` (openclaw is a separate npm sub-package whose build output is conventionally `dist/`, gitignored, regenerated by `npm run build`). All three describe-loops that iterate over `AGENTS` (worker bundle shipped, UPDATE-on-skills anti-pattern, legacy state-dir migration) now cover openclaw too. Test count: 19 → 21.

The dedicated test on `openclaw/dist/index.js` (the inlined `migrateOpenclawSkillifyLegacyStateDir` helper) stays separate — it's checking a different bundle with a different helper shape.

kaghni (approved):

PR #124, opened 2026-05-12
openclaw: patch plugins.allow at install time + drop blocking auto-recall

### LINKED ISSUE #121: OpenClaw installer doesn't patch plugins.allow → plugin loads to disk but never registers; auto-capture silently no-ops
## Symptom

User installs `@deeplake/hivemind` into OpenClaw. Files land at `~/.openclaw/extensions/hivemind/` correctly. But:

- `openclaw plugins list` does **not** show `[plugins] Hivemind plugin registered`
- Auto-capture silently does nothing — no errors, no logs, just empty sessions in Deeplake
- `/hivemind_setup` is unreachable (the agent can't see the command, because the plugin never registered)

## Root cause

OpenClaw gates plugin loading on **two** allowlists in `~/.openclaw/openclaw.json`:

1. `plugins.allow` — controls whether the plugin is allowed to load at all
2. `tools.alsoAllow` — controls whether the plugin's tools are exposed to the agent once loaded

If `plugins.allow` is an explicit non-empty array (the default-restrictive openclaw config), and `"hivemind"` is not in it, the plugin is **never loaded** — `agent_end` never fires for it, so auto-capture is a no-op.

Hivemind's installer + setup helpers only handle `tools.alsoAllow`:

- `hivemind/src/cli/install-openclaw.ts:8-30` (`installOpenclaw`) drops files on disk and never touches `openclaw.json`.
- `hivemind/openclaw/src/setup-config.ts:33-67` (`ensureHivemindAllowlisted`) and `:129-139` (`detectAllowlistMissing`) only read/write `tools.alsoAllow`.

This makes `/hivemind_setup` chicken-and-egg: the user has to add `"hivemind"` to `plugins.allow` manually before the slash command is even reachable.

## Manual fix that works (reported by a user 2026-05-11)

Edit `~/.openclaw/openclaw.json`:

\`\`\`json
{
  \"plugins\": { \"allow\": [\"...\", \"hivemind\"] },
  \"tools\":   { \"alsoAllow\": [\"...\", \"hivemind\"] }
}
\`\`\`

Then `systemctl --user restart openclaw-gateway.service`. \`openclaw plugins list\` now shows the plugin registered. **No backfill** — capture starts on the next turn.

## Fix (proposed for this session)

1. **`install-openclaw.ts`** — after copying files on disk, patch `~/.openclaw/openclaw.json` using openclaw's own allowlist semantics:
   - If `plugins.allow` is an explicit non-empty array missing `\"hivemind\"`, append.
   - If absent / empty, leave alone (don't silently flip the user from default-allow to explicit-allowlist).
   - Same for `tools.alsoAllow`.
   - Atomic write (tmp + rename) with timestamped backup.
   - Print restart hint and \"capture starts next turn — no backfill\" caveat in installer output.

2. **`openclaw/src/setup-config.ts`** — extend `ensureHivemindAllowlisted` + `detectAllowlistMissing` to also handle `plugins.allow`. Same safe semantics: only patch when it's already an explicit array.

3. **Tests** — cover all edge cases:
   - `plugins.allow` array missing hivemind → patched
   - `plugins.allow` array containing hivemind → idempotent
   - `plugins.allow` absent → untouched
   - `plugins.allow` empty array `[]` → untouched (default-allow semantics)
   - Both arrays missing → only patch what exists
   - Restart hint present in installer success message

4. **Real-world E2E** — install on a real OpenClaw gateway, send a telegram message, verify the captured turn lands in the Deeplake sessions table.

## References

- OpenClaw's own allowlist helper (the safe semantics to mirror): \`ext/openclaw/src/config/plugins-allowlist.ts:7\` \`ensurePluginAllowlisted(cfg, pluginId)\`.
- Existing setup-config: \`hivemind/openclaw/src/setup-config.ts:33-67, 129-139\`.
- Existing installer: \`hivemind/src/cli/install-openclaw.ts:8-30\`.
- Existing setup-command tests: \`hivemind/openclaw/tests/setup-command.test.ts\` (covers only \`tools.alsoAllow\` today).

## Test plan

- [ ] Unit tests for both new branches of \`ensureHivemindAllowlisted\` and \`detectAllowlistMissing\` (idempotent, preserves unrelated keys, atomic w/ backup)
- [ ] Unit tests for the installer's new openclaw.json patcher (every edge case above)
- [ ] Source review: confirm no other code path needs the same patch
- [ ] Local E2E: install onto a fixture openclaw config, verify both arrays patched, backup written
- [ ] Real-world E2E on the user's openclaw gateway via telegram channel, end-to-end:
  1. Strip the manual fix (or use a fresh config)
  2. Run the new installer
  3. Restart gateway
  4. Confirm \`openclaw plugins list\` shows Hivemind registered
  5. Send a message through telegram
  6. Confirm the turn lands in the Deeplake sessions table

\`Confidence: high\` — the root cause is mechanical (missing config patch), and openclaw already exposes the correct semantics for us to mirror.

kaghni:
Fixes #121.

## Why

Two related bugs surfaced in a user report (the original symptom: "openclaw plugin installed but auto-capture silently does nothing"):

1. **`plugins.allow` gating** — the installer copies files into `~/.openclaw/extensions/hivemind/` but never patches `~/.openclaw/openclaw.json`. If the gateway's `plugins.allow` is an explicit array missing `"hivemind"`, the plugin never registers, `agent_end` never fires, and `/hivemind_setup` is unreachable from inside the agent (chicken-and-egg: the slash command needs the plugin to load first).
2. **Blocking auto-recall** — `before_agent_start` ran a Deeplake `sessions`-table search on **every** user turn. Sessions queries currently take **11.7s / 10.8s / 11.2s** (three sequential probes) against a 10s SDK timeout, so every turn paid the full timeout before the agent could read the message. Other agents (claude-code/codex/cursor/hermes/pi) don't do this — they intercept the agent's `Grep` tool and let the agent decide when to search.

### `plugins.allow` patch in the installer

- `src/cli/install-openclaw.ts` now calls the same `ensureHivemindAllowlisted` helper that `/hivemind_setup` uses
- Mirrors openclaw's own `ensurePluginAllowlisted` semantics (`ext/openclaw/src/config/plugins-allowlist.ts:7`): only patches `plugins.allow` when it's an explicit non-empty array. Absent / empty → left alone (default-allow stays default-allow; we don't flip the user into restrictive mode and break their other plugins)
- Same logic for `tools.alsoAllow`
- Atomic write (tmp + rename) with timestamped backup
- Installer prints: which arrays were patched, restart hint, and a **"capture starts on the next turn — no backfill"** caveat so users aren't confused when older turns don't appear

### Blocking auto-recall removed

- `openclaw/src/index.ts:before_agent_start` no longer runs `searchDeeplakeTables`
- Recall flows exclusively through the agent-callable `hivemind_search` / `hivemind_read` / `hivemind_index` tools (already registered), with the `SKILL.md` body in the system prompt directing the agent to use them
- Side-effect win: the recursive `<recalled-memories>` capture bug (old recall context was getting captured as part of the user's prompt and stored) is gone too
- The hook still handles the login nudge and post-auth welcome banner — those don't touch Deeplake
- `extractKeywords` + `RECALL_STOPWORDS` removed as dead code

## Tests

- `openclaw/tests/setup-command.test.ts` — **+9 cases** for `plugins.allow` patching: idempotency, default-allow semantics, both-arrays patched, `detectAllowlistMissing` extension
- `openclaw/tests/install-openclaw.test.ts` (**new, 12 cases**) — drives the real installer against tmpdir fixtures: file copy, both patches, idempotency, default-allow preservation, malformed config tolerance, restart hint, no-duplicate
- `openclaw/tests/auto-recall.test.ts` — **rewritten**: asserts no Deeplake call on a normal turn, no `<recalled-memories>` injection, tools still registered, login-nudge path preserved

All 2214 hivemind tests pass.

## Real-world E2E on a live openclaw gateway (2026.4.21)

- Bundle installed → `Hivemind plugin registered` in gateway log
- Idempotency: re-running installer logs `allowlist already covers hivemind`, no extra backup
- **Zero `Auto-recall failed` log lines** after the restart with the new bundle (previous bundle emitted them on every turn that hit Deeplake slowness)
- `hivemind_search` / `hivemind_read` / `hivemind_index` tools confirmed registered in `before_prompt_build`

## Caveats

- **Capture (write path)** still hits the same Deeplake timeout intermittently — `Auto-capture failed: Query timeout after 10000ms` lines appear in the gateway log. `agent_end` runs *after* the agent responds, so this doesn't block the user, but it does drop some turns from memory until Deeplake's `sessions`-table query latency improves. **Separate issue, not addressed here.**
- The exact host state the original reporter hit (where adding `"hivemind"` to `plugins.allow` was needed even though tools were allowlisted) can't be reproduced on openclaw 2026.4.21 — that version has an "auto-promote plugins when their tools are in `alsoAllow`" feature that masks the gating. Older openclaw versions without that feature are the broken-state case the installer's `plugins.allow` patch is for.

## Confidence

`Confidence: high`, because the unit tests cover every edge case of the new config-patching logic, the local fixture E2E drove the real installer end-to-end against multiple fixture shapes, and the live gateway confirms the bundle loads + the recall hook no longer fires.

`Untested: openclaw versions older than 2026.4.21` (no host available; the installer's plugins.allow patch is the right fix for them but unverified end-to-end on those versions).

## Test plan

- [ ] CI green on this branch
- [ ] Reviewer reads through `openclaw/src/setup-config.ts` semantics (esp. `isPluginsAllowMissingHivemind` only-patch-explicit-arrays rule)
- [ ] Reviewer skims `openclaw/tests/install-openclaw.test.ts` and confirms test coverage matches the design intent in the description
- [ ] After merge: cut a release and verify users on default-restrictive openclaw configs see the plugin register after a fresh install

## Summary by CodeRabbit

* **New Features**
  * OpenClaw installation now validates and patches configuration to ensure the hivemind plugin is properly allowlisted, with backup creation before changes.
  * Setup command displays which allowlist fields were updated during configuration patching.

* **Changes**
  * Removed automatic memory recall that occurred on every turn; recall is now available only through direct tool invocation.
  * Improved installation and cleanup processes for extension management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/124)

### DIFF
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -833,7 +833,7 @@ export default definePluginEntry({
             if (result.delta.toolsAlsoAllow) touched.push(`"hivemind" → tools.alsoAllow`);
             return { text: `✅ Added:\n  • ${touched.join("\n  • ")}\n\nOpenclaw will detect the config change and restart. On the next turn, the agent will have access to hivemind_search, hivemind_read, and hivemind_index. **Capture starts on the next turn — earlier turns are NOT backfilled.**\n\nBackup of previous config: ${result.backupPath}${skillifyHint}` };
           }
-          return { text: `⚠️ Could not update allowlist: ${result.error}\n\nManual fix: open ${result.configPath} and add "hivemind" to BOTH the "allow" array under "plugins" AND the "alsoAllow" array under "tools".` };
+          return { text: `⚠️ Could not update allowlist: ${result.error}\n\nManual fix: open ${result.configPath}. If \`plugins.allow\` exists as a non-empty array, add "hivemind" to it. If \`tools.alsoAllow\` exists as a non-empty array, add "hivemind" to it. If either is absent or empty, leave it as-is (openclaw treats that as default-allow).` };
         },
       });
 
@@ -1183,30 +1183,33 @@ export default definePluginEntry({
     // the SKILL.md body in the system prompt directing it to call them
     // first. See issue #121 for the original report (plugins.allow gating
     // also fixed in the same PR).
-    if (config.autoRecall !== false) {
-      hook("before_agent_start", async (event: { prompt?: string }) => {
-        if (!event.prompt || event.prompt.length < 5) return;
-        try {
-          const dl = await getApi();
+    // No `config.autoRecall` gate here: the hook body no longer does any
+    // recall (CodeRabbit on #124 caught this). Both remaining paths — the
+    // login URL nudge and the post-auth welcome banner — must run for
+    // every user, including those who set autoRecall=false. Gating the
+    // whole hook registration would silently break their auth flow.
+    hook("before_agent_start", async (event: { prompt?: string }) => {
+      if (!event.prompt || event.prompt.length < 5) return;
+      try {
+        const dl = await getApi();
 
-          if (!dl && authUrl) {
-            return {
-              prependContext: `\n\nIMPORTANT: The user needs to authenticate with Hivemind. Include this link in your reply: ${authUrl}\nTell them to click it and sign in, then send another message.\n`,
-            };
-          }
-          if (!dl) return;
-
-          if (justAuthenticated) {
-            justAuthenticated = false;
-            const creds = await loadCredentials();
-            const orgName = creds?.orgName ?? creds?.orgId ?? "unknown";
-            return { prependContext: `\n\n🐝 Welcome to Hivemind!\n\nCurrent org: ${orgName}\n\nYour agents now share memory across sessions, teammates, and machines.\n\nGet started:\n1. Verify sync: spin up multiple sessions and confirm agents share context\n2. Invite a teammate: ask the agent to add them over email\n3. Switch orgs: ask the agent to list or switch your organizations\n\nOne brain for every agent on your team.\n` };
-          }
-        } catch (err) {
-          logger.error(`before_agent_start failed: ${err instanceof Error ? err.message : String(err)}`);
+        if (!dl && authUrl) {
+          return {
+            prependContext: `\n\nIMPORTANT: The user needs to authenticate with Hivemind. Include this link in your reply: ${authUrl}\nTell them to click it and sign in, then send another message.\n`,
+          };
         }
-      });
-    }
+        if (!dl) return;
+
+        if (justAuthenticated) {
+          justAuthenticated = false;
+          const creds = await loadCredentials();
+          const orgName = creds?.orgName ?? creds?.orgId ?? "unknown";
+          return { prependContext: `\n\n🐝 Welcome to Hivemind!\n\nCurrent org: ${orgName}\n\nYour agents now share memory across sessions, teammates, and machines.\n\nGet started:\n1. Verify sync: spin up multiple sessions and confirm agents share context\n2. Invite a teammate: ask the agent to add them over email\n3. Switch orgs: ask the agent to list or switch your organizations\n\nOne brain for every agent on your team.\n` };
+        }
+      } catch (err) {
+        logger.error(`before_agent_start failed: ${err instanceof Error ? err.message : String(err)}`);
+      }
+    });
 
     // Auto-capture: store new messages in sessions table (same format as CC capture.ts)
     if (config.autoCapture !== false) {
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -636,29 +636,6 @@ function buildSessionPath(config: { userName: string; orgName: string; workspace
   return `/sessions/${config.userName}/${config.userName}_${config.orgName}_${config.workspaceId}_${sessionId}.jsonl`;
 }
 
-const RECALL_STOPWORDS = new Set([
-  "the","and","for","are","but","not","you","all","can","had","her","was","one",
-  "our","out","has","have","what","does","like","with","this","that","from","they",
-  "been","will","more","when","who","how","its","into","some","than","them","these",
-  "then","your","just","about","would","could","should","where","which","there",
-  "their","being","each","other",
-]);
-
-/**
- * Extract the signal-bearing tokens from a natural-language prompt so we can
- * feed them into `searchDeeplakeTables` as a multi-word ILIKE. Mirrors the
- * pattern used by claude-code/codex grep intercepts — lowercase, strip
- * non-alphanumeric, drop short words + stopwords, cap at 4 so the SQL doesn't
- * turn into a 20-way OR.
- */
-function extractKeywords(prompt: string): string[] {
-  return prompt.toLowerCase()
-    .replace(/[^a-z0-9\s]/g, " ")
-    .split(/\s+/)
-    .filter(w => w.length >= 3 && !RECALL_STOPWORDS.has(w))
-    .slice(0, 4);
-}
-
 /** Trim a path filter down to a safe virtual prefix. `/` ⇒ unfiltered. */
 function normalizeVirtualPath(p: string | undefined | null): string {
   if (!p || typeof p !== "string") return "/";
@@ -851,9 +828,12 @@ export default definePluginEntry({
             return { text: `✅ Hivemind tools are already enabled in your allowlist.\n\nNo changes needed — memory tools are available to the agent.${skillifyHint}` };
           }
           if (result.status === "added") {
-            return { text: `✅ Added "hivemind" to your tool allowlist.\n\nOpenclaw will detect the config change and restart. On the next turn, the agent will have access to hivemind_search, hivemind_read, and hivemind_index.\n\nBackup of previous config: ${result.backupPath}${skillifyHint}` };
+            const touched: string[] = [];
+            if (result.delta.pluginsAllow) touched.push(`"hivemind" → plugins.allow`);
+            if (result.delta.toolsAlsoAllow) touched.push(`"hivemind" → tools.alsoAllow`);
+            return { text: `✅ Added:\n  • ${touched.join("\n  • ")}\n\nOpenclaw will detect the config change and restart. On the next turn, the agent will have access to hivemind_search, hivemind_read, and hivemind_index. **Capture starts on the next turn — earlier turns are NOT backfilled.**\n\nBackup of previous config: ${result.backupPath}${skillifyHint}` };
           }
-          return { text: `⚠️ Could not update allowlist: ${result.error}\n\nManual fix: open ${result.configPath} and add "hivemind" to the "alsoAllow" array under "tools".` };
+          return { text: `⚠️ Could not update allowlist: ${result.error}\n\nManual fix: open ${result.configPath} and add "hivemind" to BOTH the "allow" array under "plugins" AND the "alsoAllow" array under "tools".` };
         },
       });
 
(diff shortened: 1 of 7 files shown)


kaghni:
## User-flow E2E: drove the real `hivemind claw install` CLI against the reporter's exact broken-state config

Reviewer asked for proof that the user-facing flow (not just the unit-test-mocked one) actually patches the allowlists. Here's the transcript.

**Fixture: reporter's broken state, on a tmp `$HOME`.** No `plugins.entries.hivemind` (so openclaw can't auto-promote), neither array contains `"hivemind"`:

```json
{
  "meta": { "lastTouchedVersion": "2026.4.14" },
  "agents": { "defaults": { "model": "anthropic/claude-haiku-4-5-20251001" } },
  "channels": { "telegram": { "enabled": true } },
  "plugins": {
    "allow": ["telegram", "anthropic", "openai", "duckduckgo"],
    "entries": {
      "telegram":  { "enabled": true },
      "anthropic": { "enabled": true }
    }
  },
  "tools": {
    "profile":   "coding",
    "alsoAllow": ["memory_store"]
  }
}
```

Before:
```
plugins.allow contains 'hivemind'? False
tools.alsoAllow contains 'hivemind'? False
plugins.entries.hivemind present? False
```

**Run the real CLI users invoke:**
```
$ HOME=/tmp/hivemind-claw-install-flow-Y4fLiK node bundle/cli.js claw install
  OpenClaw       installed -> /tmp/hivemind-claw-install-flow-Y4fLiK/.openclaw/extensions/hivemind
  OpenClaw       patched plugins.allow + tools.alsoAllow in /tmp/hivemind-claw-install-flow-Y4fLiK/.openclaw/openclaw.json
  OpenClaw       backup: /tmp/hivemind-claw-install-flow-Y4fLiK/.openclaw/openclaw.json.bak-hivemind-1778620171180
  OpenClaw       restart the gateway to activate: systemctl --user restart openclaw-gateway.service
  OpenClaw       capture starts on the NEXT turn — earlier turns are NOT backfilled
```

**After:**
```json
{
  "meta":     { "lastTouchedVersion": "2026.4.14" },
  "agents":   { "defaults": { "model": "anthropic/claude-haiku-4-5-20251001" } },
  "channels": { "telegram": { "enabled": true } },
  "plugins":  {
    "allow":   ["telegram", "anthropic", "openai", "duckduckgo", "hivemind"],
    "entries": { "telegram": { "enabled": true }, "anthropic": { "enabled": true } }
  },
  "tools":    {
    "profile":   "coding",
    "alsoAllow": ["memory_store", "hivemind"]
  }
}
```

**Assertions:**
- `plugins.allow` contains `"hivemind"` ✅
- `tools.alsoAllow` contains `"hivemind"` ✅
- Unrelated keys preserved (`meta`, `agents.defaults.model`, `channels.telegram.enabled`, `plugins.entries.{telegram,anthropic}`) ✅
- Backup file written: `openclaw.json.bak-hivemind-1778620171180` ✅
- Plugin files on disk at `.openclaw/extensions/hivemind/`: `.hivemind_version`, `dist/`, `openclaw.plugin.json`, `package.json`, `skills/` ✅

**Idempotency — re-running on the same fixture:**
```
$ HOME=/tmp/... node bundle/cli.js claw install
  OpenClaw       installed -> .openclaw/extensions/hivemind
  OpenClaw       allowlist already covers hivemind in .openclaw/openclaw.json
```
Backup file count after re-run: **1** (unchanged — no spurious second backup). ✅

Before this fix, the same `hivemind claw install` against the same starting config would have left `openclaw.json` untouched: files would land on disk under `extensions/hivemind/`, but the gateway would refuse to load the plugin (`plugins.entries.hivemind: plugin disabled (not in allowlist) but config is present` per openclaw's own loader warning), so auto-capture would silently no-op and `/hivemind_setup` would be unreachable from inside the agent. That's exactly the symptom in #121.

kaghni:
## Rebased onto current `main` (0.7.26)

Force-pushed `a1ca705`. Conflicts resolved:
- Test files relocated to `tests/openclaw/` (per `21df218 chore(tests): relocate agent-specific tests to tests/<agent>/`) — `auto-recall.test.ts` and `install-openclaw.test.ts` moved, import paths adjusted to `../../openclaw/src/...`
- `bundle/cli.js` rebuilt from the merged sources
- `setup-command.test.ts` — same path adjustment for `detectAllowlistMissing` imports

Verified after rebase:
- `npm run typecheck` — clean
- `npm run build` — clean
- `npm test` — **2402/2402 passing** (including the 34 cases this PR adds)
- `npm run audit:openclaw` — still **5 CRITICAL / 2 WARN**, but those are pre-existing (the openclaw bundle has been failing the ClawHub static scan since the skillify worker landed, unrelated to this PR — filed separately as #169 with full diagnosis and proposed fix)

## Related but out of scope: #169

While verifying this rebase against the live gateway, I noticed the user's gateway had **auto-updated to 0.7.26**, replacing my branch's bundle. Investigating further found that ClawHub has already removed the hivemind plugin from its store ("This plugin does not exist or has been removed") because the openclaw bundle fails the static scan. That's tracked at #169 — separate PR, not gated on this one.

kaghni:
Thanks @coderabbitai — addressed 5 of the 6 findings on the latest push (7451e75 + bundle regen on 3b2f6cc). One left as intentional.

**Fixed:**

| # | Finding | Action |
|---|---|---|
| 1 | `before_agent_start` wrongly gated by `config.autoRecall` | Removed the gate. Hook always registers; body is auth/welcome only. Users with `autoRecall: false` would otherwise have lost the login URL nudge AND post-auth welcome banner. |
| 2 | `tools.alsoAllow` coercion to `[]` always patched | Changed to explicit-non-empty-only contract (matches `plugins.allow`). Absent/empty → leave alone. Test updated from "creates `[\"hivemind\"]`" to "stays undefined". Added a new test for the empty-array case. |
| 3 | `JSON.parse` null-check | Added in both `ensureHivemindAllowlisted` and `detectAllowlistMissing`. Returns structured `"openclaw config is not a JSON object"` error. Added a malformed-JSON test. |
| 4 | Installer silent on patch-write failure | Kept the silent path for the common "config file not found" case (gateway never started), but surface other errors via `warn()`. |
| 5 | Manual-fix message said "add to BOTH arrays" unconditionally | Reworded to be conditional on each array existing as a non-empty list. |
| 6 | `.resolves.not.toThrow()` invalid for `undefined`-resolving promise | Switched to `.resolves.toBeUndefined()` in `tests/openclaw/auto-recall.test.ts`. |

**Not changed:**

- The Hermes hook command path quoting (`bundle/cli.js` line ~3?) is unrelated to this PR's scope (plugins.allow + auto-recall). Worth filing as its own issue if it's a real concern — recommend doing so against `src/cli/install-hermes.ts` rather than fixing in the regenerated bundle file.

2404/2404 tests pass after the changes. Typecheck clean. Bundle regenerated.

openclaw/src/index.ts
```diff
@@ -1008,76 +988,51 @@ export default definePluginEntry({
...
-            fixedString: true,
-          };
-          const searchOpts = buildGrepSearchOptions(grepParams, "/");
-          searchOpts.limit = 10;
-          const rows = await searchDeeplakeTables(dl, memoryTable, sessionsTable, searchOpts);
-          if (!rows.length) return;
-
-          const recalled = rows
-            .map(r => {
-              const body = normalizeContent(r.path, r.content);
```
efenocchi, line 1234: is this part removed?

kaghni: @efenocchi yes — the entire `before_agent_start` recall block is gone. The hook handler now only runs two narrow paths: the login URL nudge (when no creds) and the post-auth welcome banner (once after device-flow). No more `extractKeywords`, no more `searchDeeplakeTables` call, no more `<recalled-memories>` injection.

Recall is now agent-initiated via the registered tools (`hivemind_search`, `hivemind_read`, `hivemind_index`) — matches how CC/Codex/Cursor/Hermes/Pi already work (their `Grep` PreToolUse intercept is the same pattern: lazy, on agent demand). The full diff for the removal is at `openclaw/src/index.ts:991-1031` on the latest commit.

Note: CodeRabbit also caught (and I fixed) that I'd left the entire hook wrapped in `if (config.autoRecall !== false)` — that would have disabled the auth nudge + welcome banner for users with `autoRecall: false`. Now unconditional.

efenocchi (approved):

Issue #110, opened 2026-05-07
openclaw: skilify spawn lock has no staleness check — abnormal worker death halts mining permanently


kaghni:
## Symptom

`openclaw/src/index.ts:tryAcquireOpenclawSkilifyLock` uses `O_CREAT | O_EXCL | O_WRONLY` to acquire a per-project worker lock. It succeeds only if the lock file does not exist, and creates an empty file on success.

There is **no staleness check.** If a worker dies abnormally — host kill, OOM, segfault, or any other path that prevents `releaseWorkerLock`'s `finally`-block from running — the empty lock file persists indefinitely, and **every subsequent `agent_end` will silently skip the spawn**. Mining stops permanently for that `project_key` until manual cleanup (`rm ~/.deeplake/state/skilify/<projectKey>.worker.lock`).

This was hit live during the 2026-05-07 E2E for PR #98: the first gate-failed worker run left a stale lock behind, and the subsequent test would have silently no-op'd had the lock not been deleted manually.

## Where

`openclaw/src/index.ts:327-335` — current implementation:

```ts
function tryAcquireOpenclawSkilifyLock(projectKey: string): boolean {
  try {
    fsMkdir(OPENCLAW_SKILIFY_STATE_DIR, { recursive: true });
    const lockPath = joinPath(OPENCLAW_SKILIFY_STATE_DIR, `${projectKey}.worker.lock`);
    const fd = fsOpen(lockPath, fsConstants.O_CREAT | fsConstants.O_EXCL | fsConstants.O_WRONLY);
    fsClose(fd);
    return true;
  } catch { return false; }
}
```

## Suggested fix

Mirror the staleness logic already implemented in `src/skilify/state.ts:tryAcquireWorkerLock` (lines 200–221):

- If the lock file exists, read its contents as a millisecond timestamp.
- If `Date.now() - parsed > maxAgeMs` (or unparseable), treat as stale → `unlink` and recreate.
- When creating a fresh lock, write `String(Date.now())` to the file (so the next run can compute age).

The OpenClaw spawn helper currently writes an empty file, so existing locks would be parsed as `NaN` and treated as immediately stale on the first patched run — clean migration, no manual cleanup required.

`maxAgeMs` should be conservative (e.g. 10 minutes) — long enough to not race a real running worker, short enough to recover within a reasonable window after a crash.

## What this followup carried over from

[...]

PR #172, opened 2026-05-15
openclaw: dedup skillify spawn per-session + stale-lock recovery (#100 + #110)

### LINKED ISSUE #100: OpenClaw skilify worker re-spawns on every agent_end (cumulative ~2-5 wasted spawns per long session)
## Summary

After PR #98 wired the skilify worker into OpenClaw's `agent_end` hook (commit `7900eb7`), a long OpenClaw session (e.g. 50 turns) ends up spawning the worker **2-5 times** instead of just **once**. Each spawn beyond the first is functionally a no-op (the worker's watermark advance prevents re-mining), but each still costs:

- ~50ms Node cold-start
- ~200ms SQL roundtrip for the watermark check query
- One `child_process` fork

…times 2-5 spawns per long session. Cumulative wasted overhead: ~1-2s per session, plus process-tree churn.

## Current behavior

`openclaw/src/index.ts` agent_end handler calls `spawnOpenclawSkilifyWorker` unconditionally on every turn. The spawn function uses a per-projectKey filesystem lock (`~/.deeplake/state/skilify/<projectKey>.worker.lock`) to prevent overlapping workers — but the lock is RELEASED when each worker exits, so subsequent agent_ends within the same session can re-acquire it and spawn fresh workers.

In a 50-turn session:
1. Turn 1 → lock acquired → real worker spawn (~30s actual mining)
2. Turns 2-X (during the ~30s) → lock held → cheap skip
3. Worker finishes → lock released
4. Turn X+1 → lock re-acquired → fresh worker spawn → 1 SQL query → exit (no work to do)
5. Repeat 3-4 every time the previous worker finishes

## Proposed fix (option B from PR #98 review)

Add a per-runtime in-memory `Set<string>` of session IDs that have already triggered a spawn:

```ts
// alongside the existing capturedCounts Map
const skilifySpawnedFor = new Set<string>();

// in agent_end handler, replace unconditional spawn with:
if (!skilifySpawnedFor.has(sid)) {
  skilifySpawnedFor.add(sid);
  try {
    spawnOpenclawSkilifyWorker({ ... });
  } catch (e: any) { ... }
}
```

~5 lines. One spawn per session lifetime per OpenClaw runtime instance. Memory cost: ~32 bytes per session ID, well below any realistic openclaw uptime.

## Trade-offs considered

- **Option A** (counter file `~/.deeplake/state/skilify/<key>.openclaw.counter` bumped on every agent_end, fire when `counter % N === 0`): rejected — ~30 lines of file I/O on every turn, persistent state across runtime restarts that nobody asked for, and the cadence is arbitrary on a single-session-at-a-time runtime
- **Option B** (in-memory Set, recommended): ~5 lines, no file I/O, robust under runtime restart (the worker's watermark state persists in `~/.deeplake/state/skilify/<projectKey>.json`, so a re-spawn after restart costs ~250ms and doesn't redo any actual mining)
- **Option C** (hybrid Set + interval): overkill for openclaw's "one session at a time" assumption

## Bundle-scan test to add

```ts
it("openclaw deduplicates skilify spawns by session_id (no re-fire within same session)", () => {
  const text = readFileSync(resolve(BUNDLE_ROOT, "openclaw", "src", "index.ts"), "utf-8");
  expect(text).toMatch(/skilifySpawnedFor\s*=\s*new Set/);
  expect(text).toMatch(/skilifySpawnedFor\.has\(sid\)/);
  expect(text).toMatch(/skilifySpawnedFor\.add\(sid\)/);
});
```

## Background

Surfaced during PR #98 reviews while validating per-agent skilify mining parity across CC / Codex / Cursor / Hermes / Pi / OpenClaw. The "fire on every agent_end with lock" pattern is functionally correct but sub-optimal — the lock prevents overlapping workers, but doesn't prevent serially redundant ones within the same session. Pi has the same surface (session_shutdown fires once at session end) but fires only ONCE per session naturally, so doesn't need this fix.

## Acceptance criteria

- One real worker spawn per OpenClaw session (verified by bundle scan + ideally a runtime test mocking spawn)
- Watermark advance still works correctly across runtime restarts (re-spawn after restart is acceptable, redundant mining is not)
- No regression on the existing per-projectKey lock (still useful for cross-process safety)

kaghni:
Fixes #100 and #110.

## Why

Two spawn-lifecycle bugs in `openclaw/src/index.ts`:

**#100 — Wasted re-spawns**: `agent_end` fires on every turn. The on-disk lock at `~/.deeplake/state/skillify/<projectKey>.worker.lock` prevents *overlapping* workers, but as soon as a worker exits and releases its lock, the NEXT `agent_end` re-acquires it and spawns a fresh worker. The fresh worker does one watermark-check SQL roundtrip, sees nothing new to mine, and exits — but each spawn costs ~50ms Node cold-start + ~200ms DB I/O. A 50-turn session ends up doing 2-5 spawns instead of 1.

**#110 — Stale locks halt mining permanently**: `tryAcquireOpenclawSkillifyLock` does `O_CREAT | O_EXCL | O_WRONLY` and treats any pre-existing lock as "live worker, skip." There's no staleness check. If a worker dies abnormally (host kill, OOM, segfault) before its `finally` releases the lock, the lock persists forever and *every* subsequent `agent_end` silently no-ops mining for that `project_key` permanently. Hit live during the 2026-05-07 PR #98 E2E — a manual `rm <lockfile>` was needed to recover.

### Per-runtime dedup (#100)

- New module-level `const skillifySpawnedFor = new Set<string>()`. Tracks which session IDs have already triggered a spawn in this gateway runtime.
- `agent_end` handler now wraps the `spawnOpenclawSkillifyWorker(...)` call in `if (!skillifySpawnedFor.has(sid)) { skillifySpawnedFor.add(sid); … }`.
- The on-disk lock stays authoritative across *processes* (e.g. multiple gateway restarts). The new in-memory Set only suppresses *within-runtime* redundancy.

### Stale-lock recovery (#110)

- Lock file now writes `String(Date.now())` on acquire (was an empty file).
- On `O_EXCL` failure, reads the existing lock body, parses it as a ms timestamp. If `Date.now() - ts > 10 minutes` OR the body is unparseable (`NaN`), the lock is treated as stale → unlinked → retry acquire.
- Mirrors the staleness logic in `src/skillify/state.ts:tryAcquireWorkerLock` for the non-openclaw agents.
- **Migration**: empty pre-existing lock files (from earlier code) parse as `NaN` and are treated as immediately stale on the first patched run — no manual cleanup needed.
- 10-minute max age is generous vs typical worker runtime (<30s + buffer). Pathological hangs longer than that release the spawn slot to the next `agent_end`, instead of leaking mining for the rest of the gateway's lifetime.

## Tests

- `npm run typecheck` — clean
- `npm test` — **2380/2380 passing** (one bundle-scan regex distance bumped 500→1500 to accommodate the new dedup comment block between `Auto-captured` and the spawn site; same assertion intent)

## Test plan after merge

- [ ] Long-running openclaw session (50+ turns). `grep -c "Auto-captured" /tmp/openclaw/openclaw-*.log` should be many; `ls ~/.deeplake/state/skillify/*.worker.lock` should show at most one mtime-bump per session (one spawn, not 2-5).
- [ ] Kill a worker mid-mine (`kill -9 $WORKER_PID`). Wait 11 minutes. Next `agent_end` should successfully re-acquire the lock (stale-recovery path).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of background worker spawning in extended agent sessions by preventing redundant spawn attempts
  * Enhanced detection and cleanup of stale worker states
  * Added error handling to gracefully manage worker startup failures

* **Tests**
  * Updated test validations for worker spawning behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/172)

### DIFF
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -68,6 +68,7 @@ import {
   existsSync as fsExists, mkdirSync as fsMkdir, openSync as fsOpen,
   closeSync as fsClose, writeFileSync as fsWriteFile, constants as fsConstants,
   readFileSync as fsReadFile, renameSync as fsRename, unlinkSync as fsUnlink,
+  statSync as fsStat,
 } from "node:fs";
 import { createHash } from "node:crypto";
 // node:child_process is stubbed in the main openclaw bundle (see esbuild.config.mjs
@@ -466,11 +467,21 @@ function tryAcquireOpenclawSkillifyLock(projectKey: string): boolean {
       return acquire();
     } catch {
       // O_EXCL failed → lock file already exists. Check staleness.
+      // There's a brief window between O_CREAT|O_EXCL and the timestamp
+      // write where a racing caller can see an empty body. Don't treat
+      // empty/NaN as immediately stale (CodeRabbit on #172) — fall back
+      // to the file's mtime to decide. If the FILE is fresh, the
+      // competitor is mid-write and we should yield; if the file is
+      // older than LOCK_MAX_AGE_MS, the previous holder crashed without
+      // writing the timestamp (or the disk lost it), and we can recycle.
       try {
         const body = fsReadFile(lockPath, "utf-8");
         const ts = Number.parseInt(body.trim(), 10);
-        const age = Date.now() - (Number.isFinite(ts) ? ts : 0);
-        if (!Number.isFinite(ts) || age > LOCK_MAX_AGE_MS) {
+        const ageByBody = Number.isFinite(ts) ? Date.now() - ts : Number.POSITIVE_INFINITY;
+        let ageByMtime = 0;
+        try { ageByMtime = Date.now() - fsStat(lockPath).mtimeMs; } catch { ageByMtime = 0; }
+        const effectiveAge = Number.isFinite(ts) ? ageByBody : ageByMtime;
+        if (effectiveAge > LOCK_MAX_AGE_MS) {
           try { fsUnlink(lockPath); } catch { /* race; recheck below */ }
           try { return acquire(); } catch { return false; }
         }
@@ -535,26 +546,36 @@ function detectOpenclawGateAgent(): GateAgent | null {
   return null;
 }
 
-function spawnOpenclawSkillifyWorker(a: OpenclawSpawnArgs): void {
+/**
+ * Returns true when the worker was actually spawned (the caller can
+ * record the session in the per-runtime dedup set). Returns false on
+ * any "didn't spawn" outcome — missing worker, no delegate gate CLI,
+ * lock not acquired, mkdir/config write failure, or spawn() throw —
+ * so the caller can let a future agent_end retry. CodeRabbit on #172
+ * caught the previous flow that recorded the session before knowing
+ * whether spawn succeeded, suppressing retries forever within the
+ * runtime.
+ */
+function spawnOpenclawSkillifyWorker(a: OpenclawSpawnArgs): boolean {
   if (!fsExists(OPENCLAW_SKILLIFY_WORKER_PATH)) {
     a.loggerWarn?.(`skillify worker missing at ${OPENCLAW_SKILLIFY_WORKER_PATH} — reinstall openclaw plugin`);
-    return;
+    return false;
   }
   const gateAgent = detectOpenclawGateAgent();
   if (!gateAgent) {
     a.loggerWarn?.(`skillify spawn: no delegate gate CLI found on PATH (need one of: claude, codex, cursor-agent, hermes, pi). Mining skipped.`);
-    return;
+    return false;
   }
   const { key: projectKey, project } = deriveOpenclawProjectKey(a.channel);
   if (!tryAcquireOpenclawSkillifyLock(projectKey)) {
     // A worker is already running for this project — skip (next agent_end may
     // re-fire after the worker releases the lock, or the worker watermark
     // advance makes the re-fire a no-op).
-    return;
+    return false;
   }
   const tmpDir = joinPath(tmpdir(), `deeplake-skillify-openclaw-${projectKey}-${Date.now()}`);
   try { fsMkdir(tmpDir, { recursive: true, mode: 0o700 }); }
-  catch (e: any) { a.loggerWarn?.(`skillify spawn: mkdir failed: ${e?.message ?? e}`); return; }
+  catch (e: any) { a.loggerWarn?.(`skillify spawn: mkdir failed: ${e?.message ?? e}`); return false; }
   const configPath = joinPath(tmpDir, "config.json");
 
   // install: "global" — openclaw has no per-project filesystem cwd, so written
@@ -595,7 +616,7 @@ function spawnOpenclawSkillifyWorker(a: OpenclawSpawnArgs): void {
     },
   };
   try { fsWriteFile(configPath, JSON.stringify(config), { mode: 0o600 }); }
-  catch (e: any) { a.loggerWarn?.(`skillify spawn: config write failed: ${e?.message ?? e}`); return; }
+  catch (e: any) { a.loggerWarn?.(`skillify spawn: config write failed: ${e?.message ?? e}`); return false; }
 
   try {
     realSpawn(process.execPath, [OPENCLAW_SKILLIFY_WORKER_PATH, configPath], {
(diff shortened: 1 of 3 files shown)


kaghni:
Thanks @coderabbitai — both findings fixed on commit 358f759.

**Fixed:**

| # | Finding | Action |
|---|---|---|
| 1 | Lock-creation race: O_CREAT|O_EXCL succeeds before timestamp is written; racing reader sees empty body → NaN → unlinks live lock → overlapping workers | When the parsed body is NaN/empty, fall back to the file's mtime (`statSync`) to compute effective age. If the FILE is fresh (mtime within `LOCK_MAX_AGE_MS`), the competitor is mid-write and we yield. Only if BOTH the body is unparseable AND the file is older than the threshold do we treat it as stale and recycle. |
| 2 | `skillifySpawnedFor.add(sid)` ran BEFORE spawn could fail — one transient failure suppressed every retry for the session within the runtime | Changed `spawnOpenclawSkillifyWorker` return type from `void` to `boolean`. All non-exception failure paths (missing worker file, no delegate CLI, lock held by fresh worker, mkdir/config write failure, spawn throw) return `false`. Only successful `realSpawn(...).unref()` returns `true`. Caller does `if (spawn(...)) skillifySpawnedFor.add(sid)` — session recorded only on successful spawn. |
| 3 (nit) | Test only checked proximity, not the actual dedup-Set primitives | Added explicit assertions in `tests/claude-code/skillify-session-start-injection.test.ts`: `const skillifySpawnedFor = new Set<string>()`, `if (!skillifySpawnedFor.has(sid))`, `skillifySpawnedFor.add(sid)`. A future refactor that drops the Set or its has/add usage will now fail this test. |

2380/2380 tests pass. Typecheck clean. No bundle drift (openclaw/dist is gitignored).

openclaw/src/index.ts
```diff
@@ -403,26 +460,36 @@ function detectOpenclawGateAgent(): GateAgent | null {
...
+ * Returns true when the worker was actually spawned (the caller can
+ * record the session in the per-runtime dedup set). Returns false on
+ * any "didn't spawn" outcome — missing worker, no delegate gate CLI,
+ * lock not acquired, mkdir/config write failure, or spawn() throw —
+ * so the caller can let a future agent_end retry. CodeRabbit on #172
+ * caught the previous flow that recorded the session before knowing
+ * whether spawn succeeded, suppressing retries forever within the
+ * runtime.
+ */
+function spawnOpenclawSkillifyWorker(a: OpenclawSpawnArgs): boolean {
```
efenocchi, line 559: why do we have the specific function?

kaghni: @efenocchi yes — openclaw has its own spawn function instead of reusing the shared `src/skillify/spawn-skillify-worker.ts`. The rationale lives in the comment block right above the openclaw spawn helper (`openclaw/src/index.ts:300-310`):

> OpenClaw can't import the shared skillify TS modules — its bundle is stubbed for child_process and code-splits the gateway. Inline the spawn shape here, keyed off the bundled sibling `skillify-worker.js`.

Two specific constraints driving the duplication:

1. **`node:child_process` is stubbed in the openclaw main bundle** by the `stub-unused-child-process` esbuild plugin (`esbuild.config.mjs:280-301`). That stub replaces `spawn/execSync/execFileSync` with no-op exports to drop CC-specific dead code (e.g. browser-opening for SSO). The shared spawn helper would inherit the stub and silently fail. Openclaw works around it by using `createRequire` to bypass esbuild and pull the real `spawn` at runtime (`openclaw/src/index.ts:78-80`).

2. **`openclaw/src/index.ts` uses code splitting** (chunks/), and the shared spawn helper transitively imports modules the openclaw chunk graph deliberately avoids. Inlining the spawn keeps the openclaw bundle's chunk topology under control.

It IS code duplication, deliberately. The two functions stay in sync via the (slightly-pun-aware) header comment `mirror of src/skillify/spawn-skillify-worker.ts` plus the shared test (`tests/claude-code/skillify-session-start-injection.test.ts` asserts both surfaces have the same wiring).

If we ever drop the `child_process` stub on the openclaw main bundle, we can DRY this — but the stub itself is also load-bearing for the ClawHub static scan (PR #170), so it'd be a meaningful architectural shift, not a quick refactor.

efenocchi (approved):

PR #182, opened 2026-05-18
embeddings: drop user-visible 'deps missing' banner, keep recycle

kaghni:
## Summary
- Strip the \`enqueueNotification({id: \"embed-deps-missing\", title: \"Hivemind embeddings disabled — deps missing\", ...})\` call from \`handleTransformersMissing()\` in \`src/embeddings/client.ts\`.
- Keep the stuck-daemon recycle (SIGTERM + sock/pid cleanup) — that's the actual self-heal, fixes the issue silently on the next call.
- Remove the now-orphaned \`_signalledMissingDeps\` flag, \`embeddingsStatus()\` user-disabled check, and \`enqueueNotification\` / \`embeddingsStatus\` imports.

## Why
The banner kept stacking on top of the primary session-start message even for users whose embeddings work correctly (the daemon recycles silently and embeddings are fine on next call). The CLI's \`embeddings status\` already documents the install command for users with persistent failures, so the banner doesn't carry unique value. Removing it reduces session-start noise without losing self-heal capability.

## Test plan
- [x] \`npm run typecheck\`
- [x] \`npm run build\`
- [x] \`npx vitest run tests/claude-code/embeddings-client.test.ts tests/claude-code/embeddings-bundle-scan.test.ts tests/claude-code/notifications.test.ts tests/claude-code/notifications-queue-lock.test.ts\` — 130/130 passing
- [x] Full suite: 2704/2705 (one unrelated flake in deeplake-fs.test.ts — confirmed pre-existing on origin/main at 40% failure rate over 5 runs)
- [ ] After merge: confirm session-start no longer shows the embeddings-disabled warning even with a known-broken daemon

## Tests pinned to the new contract
- \`embeddings-client.test.ts\`: four cases in \"transformers-missing handling\" flipped to assert \`enqueueNotificationMock\` NEVER fires
- \`embeddings-bundle-scan.test.ts\`: scan flipped from \"capture.js carries embed-deps-missing\" to \"capture.js does NOT carry embed-deps-missing\" — guards against accidental reintroduction
- Queue tests using \`embed-deps-missing\` as a fixture id switched to neutral \`dedup-fixture\` (those tests validate queue dedup, not embeddings-specific behavior)

## Summary by CodeRabbit

* **Bug Fixes**
  * Removed unnecessary user notifications about missing embeddings dependencies; the system now silently manages daemon recovery without disrupting workflows.

* **Chores**
  * Updated internal daemon lifecycle management and logging infrastructure across multiple bundles for improved reliability.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/182?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/bundle/cli.js
+++ b/bundle/cli.js
@@ -4475,13 +4475,13 @@ function maybeSignalBalanceExhausted(status, bodyText) {
     return;
   _signalledBalanceExhausted = true;
   log4(`balance exhausted \u2014 enqueuing session-start banner (body=${bodyText.slice(0, 120)})`);
-  const date = (/* @__PURE__ */ new Date()).toISOString().slice(0, 10);
   enqueueNotification({
     id: "balance-exhausted",
     severity: "warn",
+    transient: true,
     title: "Hivemind credits exhausted \u2014 top up to keep capturing",
     body: `Sessions are not being saved and memory recall is returning empty. Top up at ${billingUrl()} to restore capture and recall.`,
-    dedupKey: { reason: "balance-zero", date }
+    dedupKey: { reason: "balance-zero" }
   }).catch((e) => {
     log4(`enqueue balance-exhausted failed: ${e instanceof Error ? e.message : String(e)}`);
   });
--- a/bundle/cli.js
+++ b/bundle/cli.js
@@ -4480,11 +4480,21 @@ function maybeSignalBalanceExhausted(status, bodyText) {
     id: "balance-exhausted",
     severity: "warn",
     title: "Hivemind credits exhausted \u2014 top up to keep capturing",
-    body: "Sessions are not being saved and memory recall is returning empty. Top up at https://app.deeplake.ai/billing to restore capture and recall.",
+    body: `Sessions are not being saved and memory recall is returning empty. Top up at ${billingUrl()} to restore capture and recall.`,
     dedupKey: { reason: "balance-zero", date }
   }).catch((e) => {
     log4(`enqueue balance-exhausted failed: ${e instanceof Error ? e.message : String(e)}`);
   });
+}
+function billingUrl() {
+  try {
+    const c = loadCredentials();
+    if (c?.orgName && c?.workspaceId) {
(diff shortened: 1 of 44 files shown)


kaghni:
## Re: the 6 CodeRabbit findings — scope note

All 6 findings are on lines from #bfc8e07 (Emanuele's *feat(embeddings): hello handshake + stuck-daemon recycle + visible signal*) — **not on code this PR changes.** This PR's scope is narrow: remove the user-visible "Hivemind embeddings disabled — deps missing" enqueueNotification call. The recycle/daemon-path/migration logic is unchanged from `bfc8e07`.

Per-finding triage:

| # | File:Line | Finding | Disposition |
|---|---|---|---|
| 1 | `capture.js:1615` | `_recycledStuckDaemon` one-shot — no retry after failed recycle | Pre-existing in `bfc8e07`. Single-hook-process scope so per-process degradation is bounded. Not addressed here. |
| 2 | `pre-tool-use.js:1666` | Unset `HIVEMIND_EMBEDDINGS` migrates to `enabled=false` | **Real UX regression** but per `src/user-config.ts:67`, this is the *documented* migration rule from `bfc8e07`. Disagreement with design, not a bug in my PR. Filing follow-up issue. |
| 3 | `codex/bundle/pre-tool-use.js:1368` | Recycle should fire on any daemon-path mismatch | Same pre-existing path; self-heals at 10-min daemon idle timeout. Not addressed here. |
| 4 | `codex/bundle/stop.js:1494` | Same as #3 in stop hook | Same. |
| 5 | `hermes/bundle/shell/deeplake-shell.js:68008` | Same one-shot recycle gate as #1 | Same. |
| 6 | `pi/bundle/wiki-worker.js:7` | Pass bundled `embed-daemon.js` explicitly | Pi installs without shared deps fall back; minor. Not addressed here. |

Leaving these to a future PR scoped to the embeddings recycle/migration area. Tagging @emanuele-fenocchi-activeloop for visibility — these are findings on your `bfc8e07` work that I'm explicitly deferring.

efenocchi (approved):

PR #183, opened 2026-05-18
fix(embeddings): pi spawn-on-miss + openclaw embedding producer (#178)

### LINKED ISSUE #178: embeddings: pi spawn-on-miss bug + make openclaw an embedding producer
Follow-up to #168. Surfaced during review by @kaghni who flagged that pi and openclaw had no/minimal changes in the embed-daemon fix PR. Investigation found two related gaps worth fixing together.

## Problem 1 — pi: documented-but-missing spawn-on-miss

\`pi/extension-source/hivemind.ts:166-175\` documents auto-spawn-on-miss:

> If the socket isn't there yet, we spawn the canonical daemon at \`~/.hivemind/embed-deps/embed-daemon.js\` (deposited by \`hivemind embeddings install\`) and wait for it to listen, mirroring the auto-spawn-on-miss logic in \`src/embeddings/client.ts\`. Subsequent agents (codex, CC, cursor, hermes, ...) connect to the SAME daemon — pi pays the cold-start cost only when it's the first user on the box.

But the implementation at lines 183-210 (\`tryEmbedOverSocket\`) only calls \`connect()\` and settles \`null\` on any error. **No spawn ever happens.** Result: if pi is the first agent to talk to the daemon on a given box (e.g. after reboot), it silently writes NULL into \`message_embedding\` until some *other* agent spawns the daemon. Same regression vector this PR #168 was opened to close, on a different code path.

## Problem 2 — openclaw doesn't produce embeddings

OpenClaw is MCP-mode (\`hivemind_*\` tool contracts) and currently writes NULL on every vector column:

- \`upsertRowSql\` (memory): \`summary_embedding = NULL\` on UPDATE, \`NULL\` literal on INSERT
- session capture (\`dist/index.js:1884\`): \`message_embedding\` omitted from the column list → defaults NULL

It already imports from \`../../src/\`, already uses \`createRequire(import.meta.url)\` to bypass the esbuild stub on \`node:child_process\` (lines 79-80), and already spawns long-lived workers (\`spawnOpenclawSkillifyWorker\` at line 406). So the capability to spawn the daemon and embed is already there — we just don't.

OpenClaw must remain a **consumer-only** if the user hasn't run \`hivemind embeddings install\` (respect the opt-in invariant set in \`src/user-config.ts\` from #168 — no 600MB transformers download as a side effect). It should become a producer only when \`~/.hivemind/embed-deps/embed-daemon.js\` is already present.

## Plan

1. **Extract** a shared spawn-on-miss state machine in \`src/embeddings/standalone-embed-client.ts\` (~80 LOC). Used by openclaw (which imports from \`src/\`). Pi can't import from \`src/\` (raw .ts ship constraint) — gets a parallel inline implementation, but the test suite covers both via the shared helper.
2. **Fix pi**: replace the broken \`tryEmbedOverSocket\` with the spawn-on-miss version. Bug fix.
3. **Wire openclaw as producer**: \`tryEmbedOverSocket\` + spawn-on-miss in \`upsertRowSql\` and the sessions INSERT.

Three focused commits, matching the repo's \"never >3 src files per commit across different layers\" rule.

## Edge case matrix (test coverage required)

| # | Scenario | Expected behavior |
|---|---|---|
| 1 | Binary \`~/.hivemind/embed-deps/embed-daemon.js\` missing | NULL silently; **no spawn attempt** (user hasn't opted in) |
| 2 | Binary present, socket missing, pidfile missing | Spawn → wait-for-socket (≤5s) → embed |
| 3 | Binary present, socket alive, daemon healthy | Connect directly → embed (happy path) |
| 4 | Socket file present but daemon dead (stale socket) | Cleanup socket+pid → spawn → embed |
| 5 | Pidfile points to dead PID | Stale → spawn |
| 6 | Pidfile points to live PID but socket absent (crash post-spawn) | Respect the PID, wait for socket; on timeout → NULL (no cross-process SIGTERM — PID reuse risk, same problem #168 just fixed in \`client.ts\`) |
| 7 | Two agents race to spawn (e.g. claude_code + openclaw concurrent first-write) | Winner via \`wx\` flag on pidfile; loser connects to winner's socket once it's up |
| 8 | Spawn fails (deps deleted between binary check and spawn) | NULL silently |
| 9 | Daemon spawned but never opens socket (startup crash) | 5s timeout → NULL |
| 10 | Embed request times out | NULL |
| 11 | Daemon old version → \`unknown op\` on embed | NULL (graceful, consistent with #168 narrative) |

## Acceptance criteria

- [ ] \`src/embeddings/standalone-embed-client.ts\` exists with unit tests covering all 11 edge cases above (mock spawn + real socket against a stub daemon)
- [ ] \`pi/extension-source/hivemind.ts\` actually spawns the daemon when the socket is absent (matches its own documentation)
- [ ] \`openclaw/src/index.ts\`: \`upsertRowSql\` writes the document embedding (not NULL) when the daemon is available; session capture INSERT includes \`message_embedding\` in column list
- [ ] E2E verification on \`test_plugin/default/sessions_test\` (never prod) showing openclaw writes produce non-NULL embeddings with semantic recall working
- [ ] Coverage targets met (per-file ≥90% on all 3 new/modified files)

## Out of scope

- Auto-installing the transformers deps (\`hivemind embeddings install\`) from openclaw. Must remain explicit opt-in per #168.
- Recycle-after-hello-mismatch logic in pi or openclaw. The shared canonical daemon binary is updated when the user reruns \`embeddings install\`; pi/openclaw just consume whatever's there.

efenocchi:
Closes #178. Follow-up to PR #168 — surfaced during review by @kaghni who flagged that pi and openclaw had no/minimal changes despite the "embeddings fully wired across agents" framing.

## What this lands

Three pieces of work, separated into focused commits per the repo's "never >3 src files in one commit across layers" rule:

### 1. `src/embeddings/standalone-embed-client.ts` + tests — `c9478ec`

New helper `tryEmbedStandalone(text, kind)` for agents that don't bundle a daemon of their own (pi extension source, openclaw plugin). Mirrors the spawn-on-miss state machine in `src/embeddings/client.ts` but stripped:

- No hello/handshake. Read-only consumers never recycle a stuck daemon; recycling is the hot-path client's job, two recycle paths would race.
- No singleton, no notification side-effects.
- No SIGTERM on a live-PID pidfile with a missing socket — same PID-reuse risk PR #168 fixed in `client.ts`.

Coverage threshold added at the `client.ts` tier (90/80/90/90).

### 2. Pi spawn-on-miss bug fix — `17f9435`

Pi's existing `embed()` called `spawn(...)` bare — no O_EXCL pidfile lock, no respect for an alive owner. Two concurrent pi turns (or pi racing another agent at SessionStart) both spawned a daemon; the second crashed on bind. The header comment block described the canonical behavior but the code didn't implement it.

Replaces both `tryEmbedOverSocket` (connect-only) and the inline spawn loop with a single spawn-on-miss state machine mirroring the shared helper. `embed()` collapses to env-check → empty-check → `tryEmbedOverSocket`.

### 3. OpenClaw embedding producer — `8d7df3d`

OpenClaw previously omitted `message_embedding` from every sessions INSERT — semantic recall on openclaw sessions was broken because every row landed NULL.

Now openclaw imports `tryEmbedStandalone` and embeds the captured message before INSERT. The helper imports `spawn` from `node:child_process` at the top level, which the openclaw esbuild config replaces with a no-op stub. Without the real spawn, the auto-spawn-on-miss fallback silently does nothing. Fix: openclaw already has `realSpawn` from `createRequire(import.meta.url)`; we inject it into the helper at module load via `_setSpawnImpl` (renamed from `_setSpawnForTesting` to reflect its two legitimate use cases — tests AND bundle environments stubbing `node:child_process`).

Bundle-scan regression guard in `tests/openclaw/openclaw-embed-bundle.test.ts` locks in: exactly one `tryEmbedStandalone` call site on the auto-capture path, `message_embedding` in the INSERT column list, `_setSpawnImpl(realSpawn)` called at module load, and no INSERT that hardcodes literal NULL.

### 4. Codex pre-merge review fixes — `bb9df97`

Pre-merge `codex review` flagged 2 P1 + 1 P2:

- **P1 #1 — Empty-pidfile race.** `openSync(path, "wx")` creates the lock file BEFORE `writeSync(pid)` lands. A second caller observing the gap saw `Number("") === 0 → null → "stale"`, unlinked, and re-opened. Both callers spawned a daemon; the second crashed on bind. Fix: `readPidFile` now returns a tristate (`number | "empty" | null`); `trySpawnDaemon` treats `"empty"` as "writer in progress, wait", never unlinks. Pi's inline version also switched from `writeFileSync(path, ...)` to `writeSync(fd, ...)` so a racing unlink can't clobber.
- **P1 #2 — Pidfile leak when spawn succeeds but daemon never opens socket.** Placeholder PID stayed in the file with our (still-alive) process PID; future callers saw a "live owner" and waited forever. Fix: new `maybeCleanupOwnPlaceholder` unlinks ONLY if pidfile still contains `process.pid`.
- **P2 — Runtime validation at the socket boundary.** Daemon JSON is untrusted at runtime even though TypeScript types claim `number[]`. Both implementations now reject any non-finite element before returning the vector.

4 new unit tests (empty pidfile = no respawn, retry-after-cleanup recovery, non-finite array → null, NaN/Infinity → null) + 3 source-level regression guards in pi.

### 5. Codex follow-up: stuck empty pidfile — `f04f00a`

Codex's second pass confirmed all 3 fixes correct but flagged a residual edge: a process SIGKILL'd exactly between `openSync(wx)` and `writeSync(pid)` leaves an empty pidfile that every subsequent caller treats as "writer in progress" — silent NULL embeddings for that uid forever. Extended `maybeCleanupOwnPlaceholder` to also unlink an empty pidfile after the spawnWaitMs (5s) timeout — orders of magnitude longer than the legitimate openSync→writeSync gap.

## 11-case edge matrix (all unit-tested)

| # | Scenario | Expected |
|---|---|---|
| 1 | Binary missing | NULL, no spawn |
| 2 | Binary present, no socket / pid | Spawn → wait → embed |
| 3 | Socket alive | Connect → embed |
| 4 | Stale socket, no daemon | Spawn (daemon unlinks on bind) |
| 5 | Dead PID in pidfile | Cleanup → spawn |
| 6 | Live PID, no socket | Wait, no SIGTERM |
| 7 | Two callers race | O_EXCL: one spawns, other waits |
| 8 | spawn() throws | NULL, pidfile rolled back |
| 9 | Daemon never opens socket | 5s timeout → NULL + cleanup |
| 10 | Embed request times out | NULL |
| 11 | Daemon returns unknown-op | NULL |

## Test plan

- [x] `npm test` — 2741 / 2741 pass (was 2733 before this branch; added 8)
- [x] `npm run build` clean
- [x] `npx tsc --noEmit` clean
- [x] `codex review` — final pass returned "No new [P1] or [P2] findings"
- [x] Per-file coverage on `src/embeddings/standalone-embed-client.ts`: 96.52% statements / 84.61% branches / 94.73% functions / 100% lines (≥ 90/80/90/90 threshold added in this PR)
- [ ] E2E on `test_plugin/default/sessions_test` (NEVER prod) — manual pre-merge step using the `/tmp/e2e-embed-check.mjs` pattern from PR #168 (socket p50=10ms, write p50=402ms, semantic recall TOP-1 @ 0.7409). Will run before merge.

## Files touched

- `src/embeddings/standalone-embed-client.ts` (new, 305 LOC)
- `tests/claude-code/standalone-embed-client.test.ts` (new, 22 tests)
- `pi/extension-source/hivemind.ts` (replaces `tryEmbedOverSocket` + `embed()` spawn logic)
- `tests/pi/pi-extension-source.test.ts` (5 new regression guards)
- `openclaw/src/index.ts` (embed call + `_setSpawnImpl` injection)
- `tests/openclaw/openclaw-embed-bundle.test.ts` (new bundle-scan)
- `tests/claude-code/skillify-session-start-injection.test.ts` (regex window bump)
- `vitest.config.ts` (coverage threshold)

## Summary by CodeRabbit

* **New Features**
  * Added message embedding to the auto-capture pipeline with automatic daemon spawning and graceful NULL fallback on failures.
  * Implemented improved daemon lifecycle management with race-condition safety and per-user isolation.

* **Tests**
  * Added comprehensive test coverage for embedding client functionality and daemon behavior.
  * Added integration tests to prevent regressions in embedding wiring.

* **Chores**
  * Updated test configuration for code coverage thresholds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/183?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


openclaw/src/index.ts
```diff
@@ -1258,9 +1277,20 @@ export default definePluginEntry({
...
+            // Embed the captured message. Returns null whenever the
+            // shared daemon isn't available (binary not installed, spawn
+            // failed, timeout, etc.) — embeddingSqlLiteral then yields
+            // the literal `NULL`, preserving today's "row lands with
+            // NULL in message_embedding" behavior on every failure mode.
+            // Real vectors land only when `hivemind embeddings install`
+            // has populated ~/.hivemind/embed-deps/embed-daemon.js, in
+            // line with the explicit-opt-in rule from src/user-config.ts.
+            const embedding = await tryEmbedStandalone(line, "document");
+            const embeddingSql = embeddingSqlLiteral(embedding);
```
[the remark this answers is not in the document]
efenocchi: Reviewed and declined for this PR — the suggestion is correct in vacuum but would break cross-agent semantic comparability.

All four sibling agents that write into the same `sessions.message_embedding` column embed `line` (the JSON envelope), not `text`:

- `src/hooks/capture.ts:146` (claude-code)
- `src/hooks/codex/capture.ts:131` + `src/hooks/codex/stop.ts:126`
- `src/hooks/cursor/capture.ts:158`
- `src/hooks/hermes/capture.ts:138`

Changing only openclaw to embed `text` would mix two different embedding distributions in the same column: a `search_query: "user wants X"` query would compare against `embed({id, type, session_id, content: "user wants X", timestamp})` for the other agents and `embed("user wants X")` for openclaw — different magnitudes, different score ranges, ranking gets unbalanced.

The only place `text` is embedded today is the wiki-worker path (`src/hooks/wiki-worker.ts:197` and its per-agent mirrors), which writes to a different column (`memory.summary_embedding`) — there the input IS just the summary text, no envelope, so `text` is correct.

If the team wants to switch the entire sessions capture path to content-only embeddings (which IS likely the right long-term call — less token noise, better recall), that's a separate PR that updates all five agents and re-embeds existing rows. Out of scope here.

efenocchi:
@coderabbitai both comments addressed:

**Major — embed `text` vs `line` (declined with rationale)**: replied inline at #discussion_r3262888587. Short version: all four sibling agents (claude-code, codex, cursor, hermes) embed `line` into the same `sessions.message_embedding` column; changing only openclaw would mix two embedding distributions in shared storage and unbalance ranking. Switching the whole capture path to content-only embeddings is a separate, codebase-wide PR.

**Nitpick — strengthen SIGTERM regression (fixed in 6cc2d9c)**: ported your suggested hardening verbatim. The test now (a) positively asserts the `process.kill(pid, 0)` liveness probe is present, then (b) strips it out and asserts no other `process.kill(...)` survives — catches the bare `process.kill(pid)` default-SIGTERM regression you flagged. Tests green (19/19 in pi suite).

pi/extension-source/hivemind.ts
```diff
@@ -164,51 +164,203 @@ async function dlQuery(creds: Creds, sql: string): Promise<unknown[]> {
...
+      writeSync(fd, String(process.pid));
+    } catch {
+      return false; // sub-ms race: another caller claimed it between our unlink and reopen
+    }
+  }
+  try {
+    // No explicit `env: process.env` — it's the spawn default, and a
+    // literal `process.env` reference in source kept in lockstep with
+    // src/embeddings/standalone-embed-client.ts (which DOES go through
+    // ClawHub static-scan from the openclaw bundle).
```
kaghni, line 257: @coderabbitai why there is clawhub static scan here in pi?

pi/extension-source/hivemind.ts
```diff
@@ -164,51 +164,203 @@ async function dlQuery(creds: Creds, sql: string): Promise<unknown[]> {
...
-        } catch { settle(null); }
-        sock.destroy();
-      }
+      if (nl === -1) return;
+      clearTimeout(timer);
+      try {
+        const resp = JSON.parse(buf.slice(0, nl));
+        // Daemon may return `{ error: "unknown op" }` from an older protocol — graceful NULL.
+        if (!Array.isArray(resp.embedding)) return settle(null);
+        // Codex P2 — JSON-over-socket is untrusted at runtime. Reject any
```
kaghni: @coderabbitai same for codex

kaghni: @efenocchi can you check this one

pi/extension-source/hivemind.ts
efenocchi: @kaghni fair callout — agreed the source shouldn't name the AI review tool that flagged the invariant. Six months from now nobody knows what "codex P1" means.

Cleaned up in `cf68f97`:

- `pi/extension-source/hivemind.ts` — removed `codex P1 #1`, `codex P1 #2`, `codex P2`, `codex's residual edge` labels. Each comment now describes the invariant directly (e.g. "duplicate-spawn race", "placeholder-leak failure mode", "JSON-over-socket is untrusted").
- `src/embeddings/standalone-embed-client.ts` — same scrub.
- Tests scrubbed in the same pass for consistency.

Kept the **ClawHub** references where they appear, because that one IS a recurring CI gate (the static scan on the openclaw bundle blocks any literal `process.env` reference in a file that also `fetch()`es). Future contributors who hit it for the first time need a pointer at the rule, not just a workaround they don't understand. Different from one-time review labels.

pi/extension-source/hivemind.ts
efenocchi: @kaghni same thread — labels "Codex P2" / "codex P2" gone in `cf68f97`. The comment at the validation site now just says "JSON-over-socket is untrusted at runtime. Reject any non-finite element ..." which is what future readers actually need: the invariant, not who flagged it.

To be explicit on your earlier question's literal text ("same for codex"): nothing in the codebase actually goes through Codex CLI in CI — those references were purely review-time artifacts in source comments. No build/test dependency on Codex, no codex hivemind agent involvement here either. The PR is just the standard 5-agent setup (claude-code/codex/cursor/hermes/pi/openclaw); "codex" as a hivemind agent name stays where it appears in lists like "CC/codex/cursor/hermes" since that's a real agent name.

kaghni: @coderabbitai are there still codex p1 p2 remaining that not yet cleaned up?

kaghni (approved):

PR #187, opened 2026-05-20
fix(cli): serialize concurrent `hivemind update` with O_EXCL lock

kaghni:
## Summary

- **Bug**: `SessionStart` hooks dispatch `hivemind update` detached on every session, twice per session (sync + async) by design. With multiple Claude Code sessions starting in the same second, 2–N concurrent `npm install -g @deeplake/hivemind@latest` calls race on npm's reify step. They all try to rename the existing install to the same deterministic backup path (`.hivemind-<hash>`), all but one fail with `ENOTEMPTY`, and the winner can still be SIGKILLed mid-extract — leaving `node_modules/` populated but `package.json` and `bundle/` missing. End-user symptom: `hivemind: command not found` after autoupdate, plus an orphan `.hivemind-<hash>` backup that blocks every subsequent `npm i -g`.
- **Fix**: non-blocking `O_EXCL` pidfile lock at `~/.deeplake/hivemind-update.lock`, owned by `runUpdate()` for the lifetime of `npm install -g` + `hivemind install --skip-auth`. Late arrivals log `another hivemind update is already running (pid=N); skipping` and exit 0. Stale-holder reclaim via `process.kill(pid, 0)`. Released on every exit path. **Not** acquired for no-op paths (up-to-date, registry-fail, npx, local-dev, unknown, dry-run) so a misbehaving caller can't block real updaters.
- **This is the exact follow-up promised in `src/hooks/shared/autoupdate.ts:32-35,:53`** — the lock that was intentionally moved out of `autoupdate.ts` (because dispatch returns instantly) but never landed in `runUpdate()`.

### Deliberately NOT changed

- Number of `autoUpdate()` dispatch sites. `autoupdate.ts:32-53` documents the double-fire as intentional; concurrency is the lock's job.
- No recency cache. `autoupdate.ts:37-54` documents this as explicitly rejected on review ("publish 13:01, users started 13:00 don't pick up until 17:00 — unacceptable").

## Observed incident

```
2026-05-19 17:39:21.184  npm install -g @deeplake/hivemind@latest  (Process A)
2026-05-19 17:39:21.214  npm install -g @deeplake/hivemind@latest  (Process B)
2026-05-19 17:39:21.602  npm install -g @deeplake/hivemind@latest  (Process C)

→ B and C: ENOTEMPTY rename ... -> .hivemind-9dx9PBcH
→ A: SIGKILLed mid-tarball-extract
→ /lib/node_modules/@deeplake/hivemind/  has node_modules/ only
[...]

PR #189, opened 2026-05-20
fix(plugin-cache-gc): refcount in-use versions via `.in_use/<pid>`

### LINKED ISSUE #188: plugin-cache-gc evicts in-use bundle versions, killing long-running sessions
## Symptom

A long-running Claude Code session pinned to plugin version `<old>` starts producing hook errors after enough later autoupdates have run:

\`\`\`
⎿  UserPromptSubmit hook error
⎿  Failed to run: Plugin directory does not exist:
    /home/ubuntu/.claude/plugins/cache/hivemind/hivemind/0.7.26
    (hivemind@hivemind — run /plugin to reinstall)
\`\`\`

The session's hook commands resolve `\${CLAUDE_PLUGIN_ROOT}` to its pinned plugin version directory. When that directory has been deleted from under it, every subsequent hook fails.

## Root cause

`src/hooks/plugin-cache-gc.ts` (runs on `SessionEnd`) keeps **current version + 2 next-newest** via `DEFAULT_KEEP_COUNT = 3` in `src/utils/plugin-cache.ts`. The decision is purely position-based — it does NOT consult the `.in_use/<pid>` files that already exist in each version directory.

A session pinned to a version older than the top-3 (e.g. session started on 0.7.26, then 0.7.27 → 0.7.28 → 0.7.29 → 0.7.30 all installed across other sessions' SessionStart hooks) gets its bundle dir evicted while still alive. Next hook invocation in that session ENOENTs.

## Reproduction

1. Start a long-running Claude Code session (let it pin plugin version A).
2. From another session / another machine that shares plugin cache, trigger ≥3 hivemind autoupdates (in practice: just be on a release-heavy day — the autoupdate path fires on every SessionStart, see PR #187).
3. Original session: send a UserPromptSubmit → hook error "Plugin directory does not exist".

Observed in production on 2026-05-20 with session pinned to 0.7.26 and current at 0.7.37.

## Proposed fix

Have `executeGc()` / `planGc()` in `src/utils/plugin-cache.ts` refcount via the existing `.in_use/<pid>` files: for each candidate-for-deletion version, scan its `.in_use/` directory, `process.kill(pid, 0)` each PID, **refuse to delete if any PID is still alive**. Stale PIDs (kill throws ESRCH) can be cleaned up as a side benefit.

Open questions:
- Cross-host `.in_use/<pid>`: if the plugin cache is on a shared FS (NFS, network home), PIDs from a different host show up here. `process.kill(<foreign-host-pid>, 0)` is meaningless. Likely fix: include host info in the `.in_use/` filename, or treat any present `.in_use/` file as live regardless. (Probably the latter — false-keep is much cheaper than false-delete.)
- How does Claude Code's plugin system populate `.in_use/`? Need to audit the writer side to make sure refcounting is reliable.

## Out of scope (separate bug)

A related but distinct issue: hooks that shell out to the `hivemind` CLI binary (e.g. `src/hooks/session-start.ts:137` → `hivemind skillify mine-local`) may invoke a binary whose version differs from the session's pinned plugin version, causing protocol-mismatch failures. Fixing this requires the in-session bundle to resolve `hivemind` to a version-pinned path, not via `PATH` lookup.

## Related

- PR #187 — fixes the upstream install-race that produces the "command not found" variant of this class of bug. This GC issue is independent: GC kills sessions even when the install was perfectly clean.

kaghni:
Closes #188.

## Summary

`planGc` kept current + N-1 newest by position alone, never reading the `.in_use/<pid>` claim files Claude Code writes for each session that has resolved `${CLAUDE_PLUGIN_ROOT}` to a given plugin version. A long-running session pinned to a version that falls out of the keep window had its bundle dir evicted under it on the next SessionEnd GC, breaking every remaining hook with `Plugin directory does not exist: …/hivemind/<pinned>`.

This PR makes `planGc` a refuse-to-delete gate for versions outside the keep window: if **any** live PID still claims the version via `.in_use/<pid>`, the version is kept. Versions inside the keep window are auto-kept and not queried, so the new gate can never widen the deletion set.

## Behavior

`isVersionInUse(versionDir)` treats a claim as live iff:
1. `kill(pid, 0)` succeeds, AND
2. On Linux, `/proc/<pid>/stat` field 22 (starttime) matches the recorded `procStart` (PID-reuse safe). macOS / no-procfs accepts `kill(pid, 0)` alone — false-keep is cheap, false-delete breaks sessions.

Falls back gracefully on: missing `.in_use/`, empty `.in_use/`, malformed JSON, missing/non-numeric `pid`, dead PID, mismatched `procStart`, legacy claims without `procStart`.

## Test plan

`tests/claude-code/plugin-cache.test.ts` adds:

- [x] `isVersionInUse`: missing `.in_use` dir → false; empty `.in_use` → false; live PID + matching `procStart` → true; dead PID → false; live PID + mismatched `procStart` (Linux only) → false (covers the PID-reuse case CLAUDE.md rule 1 calls out); malformed JSON ignored; missing/non-numeric pid ignored; any-live-wins among mixed claims; legacy claim without `procStart` trusts `kill(pid, 0)` alone.
- [x] `planGc`: keeps an out-of-window version when in-use; still deletes when not in-use; does NOT call `isInUse` for versions already inside the keep window (count-based assertion per CLAUDE.md rule 6 — a buggy impl can't widen the deletion set); default real-disk `isInUse` returns false on tmpdirs without `.in_use/` (proves existing tests' behavior is preserved unchanged).
- [x] focused suite: 43/43 pass
- [x] full suite: 2814/2815 pass (the one fail is a pre-existing flake in `deeplake-fs.test.ts > batches and flushes`, unrelated and not introduced here)

## Related

- #187 — concurrent `hivemind update` lock. That PR prevents partial installs that break the `hivemind` binary itself. This PR prevents the GC from evicting bundle dirs out from under live sessions. Both are independent failure modes of "hooks fail after autoupdate" — #187 fixes the install-corruption variant, this PR fixes the GC-eviction variant.

## Known limitation

If `~/.claude/plugins/cache/` is on a shared filesystem (NFS), `.in_use/<pid>` files written by sessions on other hosts will be checked against the local kernel's PID table — false-deletes are possible. Mitigation would require including a host identifier in claim files; that's a Claude-Code-side change, not a hivemind-side one. Filing a separate issue if anyone hits this in practice.

## Release Notes

* **New Features**
  * Plugin versions in use are now automatically detected and protected from deletion during garbage collection operations.
  * Cache system enhanced with process lifecycle awareness to track version usage across different platforms.

* **Tests**
  * Added extensive test coverage for version-in-use tracking, process validation, and garbage collection decision-making.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/189?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/utils/plugin-cache.ts
+++ b/src/utils/plugin-cache.ts
@@ -1,9 +1,10 @@
 import { cpSync, existsSync, readdirSync, readFileSync, renameSync, rmSync, statSync } from "node:fs";
 import { basename, dirname, join, resolve, sep } from "node:path";
-import { homedir } from "node:os";
+import { homedir, platform } from "node:os";
 
 const SEMVER_RE = /^\d+\.\d+\.\d+$/;
 const KEEP_RE = /\.keep-(\d+)$/;
+const IN_USE_DIR = ".in_use";
 
 export function isSemver(name: string): boolean {
   return SEMVER_RE.test(name);
@@ -51,6 +52,85 @@ function isPidAlive(pid: number): boolean {
   } catch (e: any) {
     return e?.code === "EPERM";
   }
+}
+
+/**
+ * Read `/proc/<pid>/stat` field 22 (start time in jiffies since boot) on
+ * Linux. Returns null when the file can't be read (process gone, macOS,
+ * weird perms) — callers fall back to PID-only liveness in that case.
+ *
+ * field 22 = "starttime" per `man 5 proc`. Stable across the process's
+ * lifetime, monotonic across reboots (it's relative to boot time but the
+ * counter only increments while a given boot is running). Combined with
+ * the PID, it uniquely identifies a process on this host across PID
+ * reuse — exactly the property Claude Code relies on when writing
+ * `.in_use/<pid>` files with shape `{"pid":N,"procStart":"<starttime>"}`.
+ */
+function readProcStart(pid: number): string | null {
+  if (platform() !== "linux") return null;
+  try {
+    const raw = readFileSync(`/proc/${pid}/stat`, "utf-8");
+    // Field 2 ("comm") is bracketed by () and may contain spaces; skip it
+    // by splitting on the LAST closing paren, then take field 20 of the
+    // remainder (field 22 of the whole line, 1-indexed).
+    const tail = raw.slice(raw.lastIndexOf(")") + 1).trim();
+    const fields = tail.split(/\s+/);
+    return fields[19] ?? null;
+  } catch {
+    return null;
+  }
+}
+
+/**
+ * A `.in_use/<pid>` file claims that PID is actively using this plugin
+ * version. We treat the claim as live iff:
+ *
+ *   1. The PID is alive on this host (`kill(pid, 0)` succeeds), AND
+ *   2. If the file includes `procStart`, it matches `/proc/<pid>/stat`
+ *      field 22 — same process, not a PID-reused-by-the-kernel reincarnation.
+ *
+ * Returns false on malformed JSON or unreadable files. Returns true on
+ * matching live process. Returns false on dead/reused PID.
+ *
+ * macOS (or anywhere `/proc` doesn't exist) skips the procStart check
+ * and accepts kill(pid, 0) as sufficient evidence. False-keep is much
+ * cheaper than false-delete: a kept old version costs disk; a deleted
+ * in-use version breaks every hook in the session that depends on it.
+ */
+function isInUseClaimLive(claimPath: string): boolean {
+  let raw: string;
+  try { raw = readFileSync(claimPath, "utf-8"); } catch { return false; }
+  let parsed: { pid?: unknown; procStart?: unknown };
+  try { parsed = JSON.parse(raw); } catch { return false; }
+  const pid = typeof parsed.pid === "number" ? parsed.pid : Number(parsed.pid);
+  if (!Number.isFinite(pid) || pid <= 0) return false;
+
+  if (!isPidAlive(pid)) return false;
+
+  if (typeof parsed.procStart === "string" && parsed.procStart.length > 0) {
+    const actual = readProcStart(pid);
+    // Only enforce match when we can read procfs. On macOS or when
+    // /proc isn't available, actual is null and we keep the claim live.
+    if (actual !== null && actual !== parsed.procStart) return false;
+  }
+  return true;
+}
+
+/**
+ * Return true if any `.in_use/<pid>` claim file in `versionDir` references
+ * a live process. Used by `planGc` to refuse-to-delete a plugin version
+ * that some still-running session has pinned.
+ *
+ * Empty / missing `.in_use/` directory → false (no live claims).
+ */
+export function isVersionInUse(versionDir: string): boolean {
+  const inUseDir = join(versionDir, IN_USE_DIR);
+  let entries: string[];
+  try { entries = readdirSync(inUseDir); } catch { return false; }
+  for (const name of entries) {
+    if (isInUseClaimLive(join(inUseDir, name))) return true;
+  }
+  return false;
 }
 
 export interface SnapshotHandle {
@@ -140,6 +220,11 @@ export interface GcPlan {
  *
  * - Keeps the current version (from the manifest) plus the next-newest
  *   versions up to `keepCount` total.
+ * - **Also keeps any older version that a live session still claims**
+ *   via a `.in_use/<pid>` file (checked through `isInUse`). Without this,
+ *   GC happily evicts the bundle dir a long-running session is pinned
+ *   to, and every subsequent hook in that session ENOENTs on
+ *   `${CLAUDE_PLUGIN_ROOT}/bundle/...`. See issue #188.
  * - Marks stale `.keep-<pid>` snapshots (dead PID) for deletion.
  * - Leaves unknown entries (non-semver, non-`.keep-*`) alone so we
  *   never touch files the installer or user put there for other reasons.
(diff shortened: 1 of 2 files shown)


kaghni:
## Real-world tests

### Test A — synthetic `/tmp` layout, source-level

Verifies `isVersionInUse` + `planGc` + `executeGc` work end-to-end on real disk (not mocks). Live claim is this test process's PID + `/proc/self/stat` field 22 as `procStart`.

```
isVersionInUse(0.7.38) = true   ← live claim on this version
isVersionInUse(0.7.37) = false  ← no claim
plan: {"keep":["0.7.40","0.7.39","0.7.38"],"deleteVersions":["0.7.37"]}
result: {"kept":["0.7.40","0.7.39","0.7.38"],"deletedVersions":["0.7.37"]}
survivors: ["0.7.38","0.7.39","0.7.40"]
```

✓ `0.7.38` (live claim, outside the `keepCount=2` window) kept by refcount.
✓ `0.7.37` (no claim, outside window) deleted.

### Test B — bundled GC subprocess against a sandboxed `$HOME`

Stronger: exercises the **actual esbuild-produced `claude-code/bundle/plugin-cache-gc.js`** (the file shipped to users) in a real subprocess. `HOME=/tmp/fake-home-...` redirects `resolveVersionedPluginDir()` away from the real `~/.claude/plugins/cache/...` so no real plugin cache is touched.

Setup:
- 5 versions (`0.7.36`–`0.7.40`) each with a copy of the fixed bundle.
- `installed_plugins.json` points at `0.7.40` as current.
- Live `.in_use/<pid>` claim on `0.7.36` (`{"pid": <this shell>, "procStart": <real /proc value>}`).

Drive: `HOME=$FAKE node $FAKE/.claude/plugins/cache/hivemind/hivemind/0.7.38/bundle/plugin-cache-gc.js`.

Result:
| version | role | want | got |
|---|---|---|---|
| `0.7.36` | live claim, outside keep-3 | **survive** | **SURVIVED** |
| `0.7.37` | no claim, outside keep-3 | delete | **deleted** |
| `0.7.38` | top-3 | keep | present |
| `0.7.39` | top-3 | keep | present |
| `0.7.40` | current | keep | present |

✓ The bundled (not just source) GC honors the in-use refcount.

Caveat on Test B: the subprocess's final `log()` call after `executeGc` threw `ENOENT` because the fake `$HOME` had no `.deeplake/` dir (`hook-debug.log` couldn't be opened). The GC's actual disk operations completed correctly before that — the crash is purely in the trailing log call, not in the GC plan execution. Real users have `~/.deeplake/` so this is a test-setup quirk, not a regression. Worth noting separately: `defaultLog` could `mkdirSync({recursive:true})` the parent before the first append, but that's a pre-existing minor robustness gap, not introduced by this PR.

`bundle/plugin-cache-gc.js` shipped 2 references to `isVersionInUse` — confirms the fix made it into the bundled artifact.

src/utils/plugin-cache.ts
```diff
@@ -53,6 +54,85 @@ function isPidAlive(pid: number): boolean {
...
+ *
+ * field 22 = "starttime" per `man 5 proc`. Stable across the process's
+ * lifetime, monotonic across reboots (it's relative to boot time but the
+ * counter only increments while a given boot is running). Combined with
+ * the PID, it uniquely identifies a process on this host across PID
+ * reuse — exactly the property Claude Code relies on when writing
+ * `.in_use/<pid>` files with shape `{"pid":N,"procStart":"<starttime>"}`.
+ */
+function readProcStart(pid: number): string | null {
+  if (platform() !== "linux") return null;
```
efenocchi, line 70: @coderabbit why?

kaghni:
## macOS behavior — intentional design, summary for future readers

To answer "should we also test on macOS?" — the code already handles it. Documenting the macOS path here so it doesn't need to be re-derived from the source comments next time someone reads this PR.

### What runs on macOS

`readProcStart()` is gated on `platform() !== "linux"`:

```ts
function readProcStart(pid: number): string | null {
  if (platform() !== "linux") return null;
  ...
}
```

So on darwin:

1. `process.kill(pid, 0)` decides liveness — works identically to Linux (returns true / throws `ESRCH`).
2. `readProcStart()` returns `null` — no procfs available.
3. `isVersionInUse` reaches the line `if (actual !== null && actual !== parsed.procStart) return false;` — `actual` is null, the procStart check is skipped, the claim is accepted as-long-as-PID-is-alive.

### The trade-off, made explicit

Linux gets PID-reuse detection (procStart match). macOS doesn't. The failure mode on macOS is therefore:

- **Theoretical risk:** a `.in_use/<pid>` file from a previous hivemind session could match a NEW unrelated process that the OS happened to reuse that PID for, causing GC to falsely retain an old version.
- **Cost of the false-keep:** ~tens of MB of disk per orphan version, eventually cleaned up the next time the same PID isn't alive when GC runs.
- **Cost of the alternative (false-delete):** GC removes a version a live session is actively running against → bundle file vanishes mid-execution → corrupted session.

False-keep is strictly the safer failure mode, so the design accepts it. This is exactly what the code comment says:

```ts
// macOS (or anywhere /proc doesn't exist) skips the procStart check
// and accepts kill(pid, 0) as sufficient evidence. False-keep is much
// cheaper than false-delete: a kept old version costs disk; a deleted
// in-use version corrupts a running session.
```

### What the existing tests cover

The "live claim is honored" test already exercises both branches via `platform() === "linux" ? <real procStart> : "0"`. On Linux it asserts strict matching; on macOS it asserts the lenient path also keeps the version. So both code paths are covered by the unit suite on whichever OS the suite runs.

### Worth a future macOS run

If we ever get reliable macOS CI, two checks that would add real signal beyond the unit tests:

1. Run the source-level suite on darwin to catch any Node-on-darwin surprises (filesystem case, kill permission quirks). Cheap.
2. Reproduce the bundled-GC subprocess test from my prior comment on macOS — exercises the actual shipped `bundle/plugin-cache-gc.js` against a `$FAKE_HOME` with a `.in_use/<pid>` claim, confirming the lenient-liveness branch works under the bundled runtime.

Neither blocks this PR. The lenient-on-macOS behavior isn't a gap — it's the documented design choice.

efenocchi (approved):

PR #204, opened 2026-05-25
fix(auth): heal token/org drift on switchOrg and session start (all 6 agents)

efenocchi:
## Summary

- **Root cause**: `switchOrg` only rewrote `creds.orgId` while reusing the long-lived API token, but the token has `organization_id` baked into its JWT claim (minted by `POST /users/me/tokens`). After a switch, `creds.orgId` and `payload.org_id` diverged silently. Code that trusts the claim — server-side checks, the `preferredOrgId` fallback in `saveCredentialsFromToken` — kept routing to the original org.
- **Fix**: `switchOrg` now re-mints against the destination org before saving. New `healDriftedOrgToken` helper auto-corrects existing drifted creds on every session start across all 6 hivemind agents (claude-code, codex, cursor, hermes, pi, openclaw).

## Commits

1. **`fix(auth)`** — `switchOrg` mint-before-save + new `healDriftedOrgToken` helper + 5 new tests (46 total).
2. **`feat(hooks)`** — wire heal into the 4 standard hook-based agents' `session-start.ts`.
3. **`feat(pi,openclaw)`** — wire heal into the 2 agents that don't fit the standard hook pattern:
   - **pi**: inline copy (raw-`.ts`, zero-deps contract — cannot import shared helpers).
   - **openclaw**: no `session_start`, so heal runs once per process inside `getApi()` behind a `driftHealAttempted` sentinel.

## Key design choices

- **Token name suffix**: `deeplake-plugin-heal-${Date.now()}` instead of date-only. Deeplake's `/users/me/tokens` rejects duplicate `(user_id, name)` with a misleading **500 INTERNAL_ERROR** (not 409); a date-only suffix would collide as soon as a second agent heals on the same day. This was caught during cross-agent testing — first pass got 500s on codex+hermes after claude-code minted with the same date-only name.
- **Heal never throws**: on mint failure (401/403/500/network), logs and returns the original creds. Session start is never blocked.
- **Mint-before-save**: a failed mint leaves `credentials.json` untouched so re-runs are safe.

## Test plan

- [x] Unit: 46/46 vitest pass, including 5 new tests covering switchOrg success/failure/no-creds + healDriftedOrgToken no-drift / missing-claim / missing-orgId / success / mint-failure.
- [x] E2E fake-token drift × 4 hook agents (claude-code, codex, cursor, hermes): all detect drift, all 401 on mint, all preserve creds.
- [x] E2E real cross-org drift × 4 hook agents (token bound to may25_2, `creds.orgId` flipped to test_plugin): all detect drift, all mint successfully, all rotate the token in credentials.json.
- [x] E2E real cross-org drift via real `claude -p --plugin-dir ./claude-code` CLI: heal log + token rotation confirmed.
- [x] Codex bundle heal verified via direct `node session-start.js` invocation (the real `codex exec --dangerously-bypass-approvals-and-sandbox` flow skips hooks by design — not a regression).
- [ ] pi end-to-end (requires Pi editor runtime).
- [ ] openclaw end-to-end (requires openclaw gateway runtime).
- [ ] CI green.

## Summary by CodeRabbit

* **New Features**
  * Added automatic credential healing that detects and corrects misaligned authentication tokens when organization IDs drift during startup.

* **Bug Fixes**
  * Fixed legacy organization switch regression by transparently refreshing credentials with organization-bound tokens, ensuring downstream operations use valid credentials.

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/204?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -302,8 +302,12 @@ let authUrl: string | null = null;
 // Heal the legacy `org switch` regression at most once per process: if
 // creds.token's JWT org_id claim differs from creds.orgId, re-mint a
 // token bound to the destination org and rewrite ~/.deeplake/credentials.json.
-// Sentinel prevents re-running on every getApi() invocation.
-let driftHealAttempted = false;
+// Promise sentinel — not a boolean — because the heal awaits I/O and a
+// boolean would let a concurrent getApi() caller see `attempted=true`
+// while the first heal was still in flight, then skip ahead and build/cache
+// the api from still-stale credentials. With a promise, the second caller
+// awaits the first's heal and reads the freshly-healed creds.
+let driftHealPromise: Promise<void> | null = null;
 // Set by the background version check in register() when a newer version is
 // available on ClawHub. Read by before_prompt_build to inject an
 // agent-facing directive nudging it to install via its own exec tool.
@@ -676,13 +680,18 @@ async function getApi(): Promise<DeeplakeApi | null> {
   // Heal token/org drift before loadConfig reads credentials.json. Heal is
   // a no-op when claim matches orgId or when the JWT carries no claim, so
   // the only added cost on the steady-state path is a base64 decode.
-  if (!driftHealAttempted) {
-    driftHealAttempted = true;
-    try {
-      const creds = await loadCredentials();
-      if (creds?.token) await healDriftedOrgToken(creds);
-    } catch { /* heal never throws; this catch is belt + braces */ }
+  // First caller initiates the promise; subsequent concurrent callers
+  // await the same promise so they all see freshly-healed creds before
+  // loadConfig() runs (see comment on driftHealPromise above).
+  if (!driftHealPromise) {
+    driftHealPromise = (async () => {
+      try {
+        const creds = await loadCredentials();
+        if (creds?.token) await healDriftedOrgToken(creds);
+      } catch { /* heal never throws; this catch is belt + braces */ }
+    })();
   }
+  await driftHealPromise;
 
   const config = await loadConfig();
   if (!config) {
--- a/openclaw/src/index.ts
+++ b/openclaw/src/index.ts
@@ -20,7 +20,7 @@ function loadSetupConfig(): Promise<SetupConfigModule> {
 // Network-only helpers stay as static imports — auth.js no longer touches fs
 // (its credential IO moved to ../../src/commands/auth-creds.js, which we load
 // lazily below so esbuild emits it as a separate chunk).
-import { requestDeviceCode, pollForToken, listOrgs, switchOrg, listWorkspaces, switchWorkspace } from "../../src/commands/auth.js";
+import { requestDeviceCode, pollForToken, listOrgs, switchOrg, listWorkspaces, switchWorkspace, healDriftedOrgToken } from "../../src/commands/auth.js";
 import { DeeplakeApi } from "../../src/deeplake-api.js";
 
 // Lazy-loaders for the fs-touching shared modules. Each becomes its own
@@ -299,6 +299,11 @@ async function checkForUpdate(logger: PluginLogger): Promise<void> {
 // --- Auth state ---
 let authPending = false;
 let authUrl: string | null = null;
+// Heal the legacy `org switch` regression at most once per process: if
+// creds.token's JWT org_id claim differs from creds.orgId, re-mint a
+// token bound to the destination org and rewrite ~/.deeplake/credentials.json.
+// Sentinel prevents re-running on every getApi() invocation.
+let driftHealAttempted = false;
 // Set by the background version check in register() when a newer version is
 // available on ClawHub. Read by before_prompt_build to inject an
 // agent-facing directive nudging it to install via its own exec tool.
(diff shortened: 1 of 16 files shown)


openclaw/src/index.ts
```diff
@@ -668,6 +673,17 @@ function normalizeVirtualPath(p: string | undefined | null): string {
...
+  // Heal token/org drift before loadConfig reads credentials.json. Heal is
+  // a no-op when claim matches orgId or when the JWT carries no claim, so
+  // the only added cost on the steady-state path is a base64 decode.
+  if (!driftHealAttempted) {
+    driftHealAttempted = true;
+    try {
+      const creds = await loadCredentials();
+      if (creds?.token) await healDriftedOrgToken(creds);
+    } catch { /* heal never throws; this catch is belt + braces */ }
+  }
```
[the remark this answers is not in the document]
efenocchi: Fixed in 4d53b726ff71ed62932675f3b7ac175cce4f46a7. Replaced the boolean sentinel with a promise sentinel (`driftHealPromise: Promise<void> | null`) — first caller into `getApi()` initiates the heal promise, all subsequent concurrent callers await the same promise, so they all read freshly-healed creds before `loadConfig()` runs. Race closed.

src/commands/auth.ts
```diff
@@ -175,7 +175,55 @@ export async function listOrgs(token: string, apiUrl = DEFAULT_API_URL): Promise
...
+  // /users/me/tokens). Re-mint against the destination org so the claim
+  // matches creds.orgId — otherwise anything that trusts the token claim
+  // instead of the X-Activeloop-Org-Id header resolves to the old org.
+  const apiUrl = creds.apiUrl ?? DEFAULT_API_URL;
+  const tokenName = `deeplake-plugin-${new Date().toISOString().slice(0, 10)}`;
+  const tokenData = await apiPost("/users/me/tokens", {
+    name: tokenName,
+    duration: 365 * 24 * 3600,
+    organization_id: orgId,
+  }, creds.token, apiUrl) as { token: { token: string } };
```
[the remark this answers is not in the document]
efenocchi: Fixed in 4d53b726ff71ed62932675f3b7ac175cce4f46a7. Switched switchOrg's token name to `deeplake-plugin-switch-${Date.now()}` for parity with the heal path — same hazard, same fix. Comment updated to point at the documented 500-on-duplicate failure mode.

openclaw/src/index.ts
```diff
@@ -668,6 +677,22 @@ function normalizeVirtualPath(p: string | undefined | null): string {
 async function getApi(): Promise<DeeplakeApi | null> {
   if (api) return api;
 
+  // Heal token/org drift before loadConfig reads credentials.json. Heal is
+  // a no-op when claim matches orgId or when the JWT carries no claim, so
+  // the only added cost on the steady-state path is a base64 decode.
+  // First caller initiates the promise; subsequent concurrent callers
+  // await the same promise so they all see freshly-healed creds before
+  // loadConfig() runs (see comment on driftHealPromise above).
+  if (!driftHealPromise) {
```
kaghni, line 686: Shouldn't we instead do the check during org switch?
Openclaw has it's separate /hivemind_switch_org command besides bash commands

pi/extension-source/hivemind.ts
```diff
@@ -93,6 +93,52 @@ function loadCreds(): Creds | null {
   }
 }
 
+// Inline copies of decodeJwtPayload + healDriftedOrgToken (the shared helpers
+// live in src/commands/auth.ts, but pi extensions ship as raw .ts with no
+// shared-module imports — kept in lockstep with that file).
+function decodeJwtPayloadInline(token: string): Record<string, unknown> | null {
```
kaghni, line 99: can't we reuse healDriftedAuthToken for pi?

openclaw/src/index.ts
efenocchi: Good question — short answer is both layers are doing different work:

- **`switchOrg` in commit 1** already re-mints on switch, so future switches (whether via `/hivemind_switch_org` in openclaw, the bash CLI `hivemind org switch`, or anywhere else) won't introduce drift.
- **The boot heal in `getApi()`** is the recovery path for users whose `credentials.json` already drifted from a *pre-fix* switch call — their token claim and `creds.orgId` are already out of sync, and they won't invoke switch again (they're already on the org they want). A switch-time-only check wouldn't reach them. The first `getApi()` call after install does.

So the in-switch fix prevents new drift, the boot heal cleans up legacy drift. Both layers stay.

pi/extension-source/hivemind.ts
efenocchi: Wanted to, but pi's extension contract blocks it. `pi/extension-source/hivemind.ts` is loaded by pi as raw TypeScript with no bundler — every import has to resolve at runtime from the pi extension dir, where `../../src/commands/auth.js` doesn't exist. The file's existing comments call this out explicitly ("raw .ts, zero deps" at L165, "no shared-module imports allowed" at L1088), and every other import in the file is a `node:*` builtin for the same reason.

Sharing would mean either (a) bundling pi the way claude-code / codex / cursor / hermes are bundled, or (b) symlinking auth.ts under pi/extension-source/. Both are bigger lifts than this PR. For now I kept the inline copy short (~30 lines) and noted in the commit message that it must stay in lockstep with `healDriftedOrgToken` in `src/commands/auth.ts` — happy to file a follow-up issue for the shared-helper unification if it's worth queueing.

kaghni: @efenocchi bundling pi is low priority issue I think can you create on issues?

kaghni (approved):

PR #223, opened 2026-05-30
First-run + resume session briefs, kept to the user channel

khustup2:
## Summary
Adds two SessionStart briefs and makes a hard rule that **every user-facing notification stays out of the model's prompt context**. The channel-safety half is prompted by a user flagging our credit notice as a prompt-injection attempt — it was.

## Briefs
- **cold-start-brief** — first-run insight mined from local history. CTA + cadence are sign-in-aware:
  - *anonymous* → a sign-in nudge that re-surfaces at most once/24h until they convert (the install→sign-in conversion slot);
  - *signed-in* → a one-time onboarding brief, after which **resume-brief** takes over.
- **resume-brief** — "where you left off" for signed-in users, from the latest project summary. Query bounded at 1.5s so it can't stall the hook; pointer skips `# Session`/metadata/`**Label**` boilerplate and returns the first prose sentence (verified against real summaries).

## Channel safety (the important part)
A user's agent flagged repeated `HIVEMIND ALERT` system-reminders as prompt-injection. They were right — it was our own `capture.ts` writing `"HIVEMIND ALERT … Tell the user clearly: … top up at <url>"` into the model's `additionalContext` on every failed capture.

- Notifications banner is now **`userVisibleOnly` across the board** — welcome / savings / briefs render to `systemMessage`, never `additionalContext`. Mined and summary-derived prose can no longer reach the model.
- `capture.ts` no longer emits the mid-session credit notice into the model channel (no safe user channel exists for PostToolUse, so we don't smuggle it via the model). The credit-exhausted condition reaches the **user** via the SessionStart balance banner.
- `balance-exhausted` / `low-balance` notifications → `userVisibleOnly`.
- `session-start` no-auth note is neutral/factual (dropped the imperative *"ask the user to run /hivemind:login"*).
- `localMinedRule` unregistered: the cold-start signup brief owns the single anonymous conversion slot — no double login nudge.

## Tests
- New channel-split regression: a resume brief carrying an injection-shaped payload reaches `systemMessage` but **never** `additionalContext`.
- New prose-extraction unit tests (boilerplate skip, mid-token-dot safety, empty-on-no-prose).
- Updated welcome/savings assertions to the user channel.
- Full `tests/claude-code` + `tests/shared` suites green (2833 tests); `tsc --noEmit` clean.

## Known follow-ups (minor, same class, not in this PR)
- `updateNotice` and `localMinedNote` are still appended to `session-start`'s `additionalContext` — milder (no "tell the user / pay"), but should move to the user-only path.
- Add a CI guard that fails if any `additionalContext` write contains billing / `hivemind login` / `hivemind:update`-style user-directed strings.

## Release Notes

* **New Features**
  * Session briefs now appear on welcome, showing insights from your last project or suggesting next steps on first run
  * Personalized greetings for first-time and returning users
  * Simplified messaging when not logged into Deeplake

* **Bug Fixes**
  * Removed disruptive mid-session balance notifications
  * Billing notices now properly handled as user-visible content only

* **Tests**
  * Added test coverage for session brief extraction and welcome banner logic

[Review Change Stack](https://app.coderabbit.ai/change-stack/activeloopai/hivemind/pull/223?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


kaghni:
Reviewed end-to-end. The client-channel fix is solid:

- All client enqueues (`balance-exhausted`, `low-balance-warning`, welcome, savings, briefs) carry `userVisibleOnly: true` — verified in source and in the built bundle.
- `emitClaudeCode` correctly splits: `modelSafe = notifications.filter(n => !n.userVisibleOnly)`, so flagged notifications only reach `systemMessage`.
- The mid-session `emitBalanceExhaustedInline` "HIVEMIND ALERT" path in `capture.ts` is gone — that was the loudest offender.
- 85/85 tests pass; CI fully green incl. the new Windows leg.

**One gap I'd flag before declaring the injection vector closed**: `src/notifications/sources/backend.ts:toClient` doesn't set `userVisibleOnly` on backend-pushed rows. So org-targeted notifications coming from `GET /me/notifications` — including the deeplake-api `MaybeWarn` row whose body reads *"Low Deeplake balance: … Top up to avoid service interruption"* — still flow to `additionalContext` (model-visible). Same prompt-injection shape, different source.

Two ways to close it:
1. **Backend (deeplake-api)**: set `Targets: []string{"email"}` on the `MaybeWarn` notification so hivemind never fetches it as a banner row.
2. **This PR**: mark backend rows `userVisibleOnly: true` in `toClient` for billing-class ids (`backend:low_balance_warning_…`).

(Disregard my earlier mention of CodeRabbit findings — I checked HEAD and the path-separator + fireReason ones are already fixed in `f575ae3f`, the threads just weren't replied to with the SHA.)

kaghni (approved):

kaghni:
Re-reviewed against `0e84c706` — the backend-push gap is closed. `toClient` now sets `userVisibleOnly: true` on every server-pushed notification, with a comment that explicitly calls out the deeplake-api low-balance "top up to avoid service interruption" push as the motivating case. 85/85 tests still green after the change. From my side this clears the residual injection vector I flagged earlier.

PR #225, opened 2026-06-02
fix(goals): stop the generic memory table from re-surfacing phantom goals in the VFS

khustup2:
## Problem
VFS↔goals-table version skew. Pre-routing hook versions (<=0.7.4) wrote goals as plain files into the **generic `memory` table**. The VFS bootstrap (`memoryBootstrap` in `deeplake-fs.ts`) loads *every* memory path, so those goal-shaped rows reappear under `/goal/<owner>/<status>/...` even though they were never in `hivemind_goals`.

Result: a goal is visible in `ls /goal/...` but **absent from `hivemind goal list`** (the CLI reads only the structured table) — and vice-versa. Confirmed live: a phantom goal surfaced in the VFS that `hivemind goal get` reported as *not found*.

## Fix
`memoryBootstrap` now skips goal/kpi-shaped paths **when the dedicated table is configured**, so that namespace is sourced solely from `hivemind_goals` / `hivemind_kpis` and the VFS and CLI agree. When routing is off (no dedicated table) the rows are the only copy, so they are kept.

## Verification (against rebuilt bundle)
- Phantom goal no longer appears in VFS — VFS and CLI now agree ✅
- VFS write → reaches `hivemind_goals` ✅
- CLI create → visible in VFS `ls` ✅
- VFS `mv` opened→closed → table status updates, no dual-listing ✅
- 2 regression tests added; full `deeplake-fs` suite (74) green; `tsc --noEmit` clean

Fixes the open goal *"Fix Hivemind VFS to goals-table version skew"*.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where legacy goal and KPI entries would incorrectly appear in listings when dedicated goal/KPI tables are configured, ensuring proper namespace isolation between generic and structured storage.

### DIFF
--- a/src/shell/deeplake-fs.ts
+++ b/src/shell/deeplake-fs.ts
@@ -224,6 +224,20 @@ export class DeeplakeFs implements IFileSystem {
         const rows = await client.query(sql);
         for (const row of rows) {
           const p = row["path"] as string;
+          // Goal/KPI-shaped paths belong exclusively to the dedicated
+          // hivemind_goals / hivemind_kpis tables. Pre-routing hook
+          // versions (<=0.7.4) wrote goals to the generic memory table
+          // as plain files; surfacing those here re-injects phantom
+          // goals into the VFS goal namespace — visible in `ls /goal/...`
+          // but absent from `hivemind goal list` (the CLI reads only the
+          // structured table). Skip them when the dedicated table is
+          // configured so the two views stay in sync. (When it is not
+          // configured, goal routing is off and these rows are the only
+          // copy, so we keep them.)
+          const kind = classifyPath(p);
+          if ((kind === "goal" && fs.goalsTable) || (kind === "kpi" && fs.kpisTable)) {
+            continue;
+          }
           fs.files.set(p, null);
           fs.meta.set(p, {
             size: Number(row["size_bytes"] ?? 0),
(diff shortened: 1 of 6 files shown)


vitest.config.ts
```diff
@@ -462,6 +462,15 @@ export default defineConfig({
...
         "src/utils/repo-identity.ts":        { statements: 85, branches: 50, functions: 90, lines: 90 },
+        // fix/goals-vfs-skew — the Deeplake VFS (goal/kpi routing, graph
+        // bridge, session reads, soft-close/status-transition). Previously
+        // unenforced: the file sat at ~82% lines / ~72% branches and only
+        // surfaced as a red PR-coverage comment. deeplake-fs-coverage.test.ts
+        // exercises the structured-table dispatch, the /graph/* bridge, the
+        // session concat path, and the rm/mv goal flows against a stateful
+        // mock DeeplakeApi, bringing it to 97/91/97/99. Floor set at 90 to
+        // catch regressions on these paths going forward.
+        "src/shell/deeplake-fs.ts":          { statements: 90, branches: 90, functions: 90, lines: 90 },
```
efenocchi, line 473: @coderabbit why did we add this?

src/shell/deeplake-fs.ts
```diff
@@ -224,6 +224,20 @@ export class DeeplakeFs implements IFileSystem {
...
+          // goals into the VFS goal namespace — visible in `ls /goal/...`
+          // but absent from `hivemind goal list` (the CLI reads only the
+          // structured table). Skip them when the dedicated table is
+          // configured so the two views stay in sync. (When it is not
+          // configured, goal routing is off and these rows are the only
+          // copy, so we keep them.)
+          const kind = classifyPath(p);
+          if ((kind === "goal" && fs.goalsTable) || (kind === "kpi" && fs.kpisTable)) {
+            continue;
+          }
```
efenocchi, line 240: @coderabbit what does it do?

efenocchi (approved):

PR #228, opened 2026-06-03
fix(graph): cloud pull sha + cross-agent VFS parity (Cursor/Codex/Hermes)

efenocchi:
## Summary

Code-graph fixes + cross-agent parity (no docs in this PR — the planning PRD is kept local only).

1. **fix(graph): cloud pull sha verification** — P1. Push stored `snapshot_sha256` over the **stable fields only** (excludes `observation`); pull was hashing the **entire payload** and comparing the two, so every real cloud row was rejected with `snapshot_sha256 mismatch` and team pull silently failed. Pull now parses the payload, validates `nodes`/`links`, and recomputes via `computeSnapshotSha256` (the same function push uses).

2. **feat(graph): cross-agent VFS parity** — the code graph previously worked only on Claude Code. New shared parser `src/graph/graph-command.ts` (`tryGraphRead`) is wired into Cursor, Codex, and Hermes pre-tool-use hooks; Cursor also gets the SessionStart graph inject and the auto-build hook (`graph-on-stop`) in its bundle + hook registration. Hardened over two rounds of Codex review (pipe-swallow, path traversal, quoted paths/flags, multi-file reads, per-agent cwd).

3. **feat(graph): `hivemind-graph` skill** for claude-code / codex / hermes (Cursor & pi have no skill loader and rely on the inject).

## Tests
- `tests/shared/graph/deeplake-pull.test.ts` (+5 regression) and `tests/shared/graph/graph-command.test.ts` (16) green.
- Full suite 3751 green; `npm run build` green. Pre-existing unrelated teardown error in `session-notifications-hook.test.ts` only.

## Session Context
[Session transcript](https://gist.github.com/efenocchi/f1e8cc197a90d218b499f2edb97d5d60)

## Summary by CodeRabbit

* **New Features**
  * Multi-language code graph: TypeScript, JavaScript, and Python support; cross-file resolution (calls/imports/heritage) and new VFS endpoints: query/, impact/, neighborhood/, layers/, tour/, path/

* **Bug Fixes**
  * Stronger cloud snapshot validation: rejects malformed or tampered payloads and verifies stable-field hashes before accepting snapshots

* **Documentation**
  * Hivemind Graph skill guides added for multiple agents

* **Tests**
  * Broad multi-language, cross-file, VFS, and regression test coverage

* **Chores**
  * Added tree-sitter-python and ignored evals/ directory


docs/GRAPH-ROADMAP.md
```diff
@@ -0,0 +1,905 @@
...
+Blocco C  query/ one-shot (2-in-1) + VFS arricchito  ← UX agente
+    ↓
+Blocco D  Ricerca fuzzy su find/query                 ← search migliore
+    ↓
+Blocco E  Enrichment LLM (async, opt-in)             ← summary, tags, layers
+    ↓
+Blocco F  Embeddings + hybrid search                 ← NL query, ranking vector
+    ↓
+Blocco G  Team enriched sync + altre lingue          ← dopo core solido
+```
```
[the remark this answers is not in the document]
efenocchi: Skipping this one: there is no markdownlint in CI for this repo (no workflow, no `package.json` script, no `.markdownlint*` config), so MD040 cannot fail any pipeline — these fenced blocks are doc-only and purely cosmetic. Leaving the roadmap as-is to keep the diff focused on the code fix.

src/graph/deeplake-pull.ts
```diff
@@ -148,11 +149,28 @@ export async function pullSnapshot(
...
+  // worktrees/branches/timestamps dedups.
+  let parsedSnapshot: GraphSnapshot;
+  try {
+    parsedSnapshot = JSON.parse(cloudPayload) as GraphSnapshot;
+  } catch (err) {
+    return errorOutcome("parse cloud snapshot", err);
+  }
+  if (!Array.isArray((parsedSnapshot as { nodes?: unknown }).nodes) ||
+      !Array.isArray((parsedSnapshot as { links?: unknown }).links)) {
+    return errorOutcome("parse cloud snapshot", new Error("snapshot missing nodes/links arrays"));
```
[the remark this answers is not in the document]
efenocchi: Fixed in 413b9f0b. Added a `parsedSnapshot === null || typeof parsedSnapshot !== "object"` guard right after `JSON.parse`, before the `nodes`/`links` access, so a payload like the literal `"null"` returns the documented `error` outcome instead of throwing `TypeError` on `null.nodes`. Covered by a new test in `tests/shared/graph/deeplake-pull.test.ts` (cloud `snapshot_jsonb` = `"null"` → error outcome, no throw, no files written).

efenocchi:
@coderabbitai review

efenocchi:
@coderabbitai review

efenocchi:
@coderabbitai — triage of the open findings, none require a code change:

**1. `src/graph/deeplake-pull.ts` 158-166 — non-null object guard before property access.**
Already addressed in `413b9f0b`. The `parsedSnapshot === null || typeof parsedSnapshot !== "object"` guard is at lines 168-170, returning the documented `errorOutcome("parse cloud snapshot", ...)` *before* any `.nodes`/`.links` access. Resolved.

**2. `src/graph/deeplake-pull.ts` 168-177 — verify payload `graph` identity matches head/repoKey (Major).**
Declining as defense-in-depth not required on this path. The SELECT already key-matches the row: `WHERE ... repo_slug = '<repoKey>' AND commit_sha = '<head>'` (lines 126-128), so the returned row belongs to the requested `(repoKey, head)`. `push` writes the payload's `graph.commit_sha`/`graph.repo_key` from those same values, and the `snapshot_sha256` check (lines 179-183) verifies payload integrity. A payload-vs-column identity mismatch would require a backend data-integrity bug, which is out of scope for this read path. Happy to add explicit equality asserts if we want belt-and-suspenders, but it's not a correctness gap.

**3. `docs/GRAPH-ROADMAP.md` line 9 (`/tmp/...` reference) and lines 159-173 (MD040 fenced-code language).**
Acknowledged nitpicks. This roadmap is slated for an English rewrite/cleanup; both the durable-link swap and the MD040 language tags will be folded into that pass.

efenocchi:
@coderabbitai review

Commit `10d22edf` converts `src/graph/resolve/cross-file.ts` from a git-binary file (it had a literal NUL byte as the dedup-key separator) to plain UTF-8 text. The full file — including the new `resolvePythonModule` / `matchPythonSuffix` logic — is now diffable and reviewable for the first time.

Please focus this review on `src/graph/resolve/cross-file.ts` only; this commit changes no other files.

efenocchi:
@coderabbitai review

Addressed your `src/graph/resolve/cross-file.ts` findings in commit `c91a77b0`:

- **🐛 Bug (Medium) — over-climbing relative import.** Fixed: `resolvePythonModule` now tracks the climb count and returns `null` when the leading dots exceed the importer's parent directories (instead of clamping to `.` and matching at repo root). Added a regression test (`....foo` from `pkg/sub/mod.py` → `null`).
- **💬 Nit (Low) — ambiguous early return.** Added a comment clarifying that ambiguity on any `matchPythonSuffix` target form is intentionally sufficient to drop the whole resolution.
- **⚡ Perf (Medium) — O(files × 4) suffix scan.** Acknowledged; deferred as a follow-up. The resolver runs once at build time and is fine at current repo sizes; the suffix-keyed reverse index is a good optimization for very large monorepos and will be tracked separately.

This commit touches only `cross-file.ts` and its test.

efenocchi:
@coderabbitai Addressed in the latest commit: the `hivemind graph` USAGE string and the file-header comment now say "TypeScript / JavaScript / Python" instead of "TypeScript only / TypeScript source files". (The `Discovered N source files` build log was already de-TypeScript'd in the ignore-config commit.) No other "TypeScript only" references remain in `src/commands/graph.ts`.

scripts/ensure-tree-sitter.mjs
```diff
@@ -17,7 +17,7 @@ import { createRequire } from 'node:module';
```
kaghni, line 17: @coderabbitai what tree-sitter is and is used for?

src/cli/install-codex.ts
```diff
@@ -29,11 +29,25 @@ function buildHooksJson(): Record<string, unknown> {
       UserPromptSubmit: [hookCmd("capture.js", 10)],
       PreToolUse: [hookCmd("pre-tool-use.js", 10, "Bash")],
       PostToolUse: [hookCmd("capture.js", 15)],
-      Stop: [hookCmd("stop.js", 30)],
+      // One Stop matcher-block with TWO commands — stop.js (capture) +
+      // graph-on-stop.js (code-graph auto-build, G3). Single block (not two)
+      // mirrors the static codex/hooks/hooks.json and keeps one entry per
+      // event for the merge/dedupe logic.
+      Stop: [stopBlockWithGraph(30)],
```
kaghni, line 36: @coderabbitai so do we fire graph on every stop? why?

src/graph/extract/python.ts
```diff
@@ -0,0 +1,394 @@
+/**
```
kaghni, line 1: @coderabbitai why we have python extracter what's the usecase for hivemind

src/graph/vfs-handler.ts
```diff
@@ -30,6 +30,11 @@ import { readLastBuild } from "./last-build.js";
 import { repoDir } from "./snapshot.js";
```
kaghni, line 30: @coderabbitai what vfs-handler is what it does what's it's usage

kaghni (commented):
@coderabbitai pi and openclaw are missing for hivemind-graph skills and corresponding graph implmentation if needed can you create an issue for that?

src/cli/install-codex.ts
kaghni: I think it can be too much is there other way closer to session end

kaghni: @efenocchi can you check if Stop hook runs on every agent assitant turn it might be too much

efenocchi: @coderabbit to keep the graph consistent maybe is better to move the logic at the stop hook in the other AI assistants as well. the price should be contained since the graph to be recomputed is just for the modified files. Am I wrong? having a different logic between different AI assistants doesn't look good to me, but I might be wrong

efenocchi: @coderabbit is it done asynchronously for all the ai assistants, right?

kaghni (approved):

PR #232, opened 2026-06-04
fix(wiki): gate Next Steps on a hard consequence test, not a base-rate guess

khustup2:
## Problem

The resume **"pick up where you left off"** brief still surfaced false-positive next steps. The wiki worker would invent resumable work for sessions that had actually wrapped clean, so the SessionStart resume banner pointed users at non-existent open work.

This is a **generation** problem, not a parsing one — the `resume-brief` parser already treats `none` as authoritative and won't fall back to stale TODOs. The leak was the wiki-worker LLM being too generous when it *writes* the `## Next Steps` section.

## Change

Tighten the generation prompt across all four agent copies (`claude` / `hermes` / `cursor` / `codex` — kept byte-identical):

- **Remove the base-rate framing** (the old "most sessions are DONE when they end"). The right frequency depends entirely on the use case — long multi-session projects legitimately have a next step most of the time — so it must not bias the judgment in either direction.
- **Require a concrete consequence gate.** A next step is warranted ONLY when omitting it would make the project *miss something important with substantial consequences*: work silently lost, a known bug/regression shipping to users, data or state left corrupted/unsafe, a security/data-integrity hole, or a critical blocker/decision a returning engineer would **not** rediscover on their own.
- **Force the model to name the specific bad outcome** before writing anything but `none`. If it can't — or a competent engineer would just re-derive the work from code/tests/git — it writes `none`.
- **Spell out the non-qualifiers**: nice-to-haves, polish/cleanup, "could also"/"consider", the obvious natural continuation, open-ended exploration, trivial/mechanical follow-ups, and all admin wrap-up (commit/push/PR/deploy/CI).

## Scope / safety

- Generation-side prompt text only. No code-path, parser, or schema changes.
- `npm run typecheck` clean; existing `resume-brief` tests (54) unchanged and still green.
- Takes effect once the plugin is rebuilt/bundled (`WIKI_PROMPT_TEMPLATE` is embedded at build time).

> Note: the goals timestamp/heal "old changes" that were in my local working tree turned out to be **already merged into `main`** (in a slightly newer form), so this PR intentionally contains only the new Next Steps work.

## Summary by CodeRabbit

* **Improvements**
  * Tightened "Next Steps" guidance so the default is exactly "none" and a single concrete imperative is emitted only when omitting it would cause a substantial, non‑obvious bad outcome—making suggestions rarer but more meaningful across agents and extensions.
* **Tests**
  * Added/expanded tests that lock the Next Steps prompt contract across templates and broaden spawn/resolve coverage, including deterministic worker resolution, fallback behaviors, and env-var propagation.


kaghni:
@coderabbitai should this be added also to openclaw and pi?

efenocchi (approved):

PR #234, opened 2026-06-05
fix(wiki): rebalance Next Steps gate — unfinished work always qualifies

khustup2:
## Problem

The stricter gate shipped in #232 over-corrected: it stopped emitting next steps even for sessions that clearly ended **mid-task**. A real example — a session whose final message explicitly says the work isn't done still produced `none`.

Two clauses caused it:
1. *"…or a competent engineer would simply re-derive the work from the code, tests, comments, or git state — write none."* — almost **any** unfinished coding work is re-derivable from git state, so this suppressed legitimate unfinished tasks.
2. The *"MISS SOMETHING IMPORTANT WITH SUBSTANTIAL CONSEQUENCES / materially harmed"* framing — ordinary "I didn't finish X" never feels catastrophic, so the model judged it below the bar.

The root mistake was collapsing **two different questions** into one catastrophe test.

## Fix — a two-step gate

- **STEP 1 — Is the session's own work finished?** If it ended mid-task (feature half-implemented, build/test still failing, fix written-but-unverified, plan not executed, blocker unresolved, explicit "still need to…"), it is **not** finished and a concrete next step is **required**. The session's **last messages** are called out as the strongest signal — directly addressing the failure mode where the final message says the work isn't done.
- **STEP 2 — If the core work IS finished**, default to `none`. The strict "separate, important, non-obvious, materially-harmed" test now applies **only** to this finished-work exception — never to unfinished work. Admin wrap-up (commit/push/PR/deploy/CI) and open-ended exploration still count as `none`.

This keeps the false-positive suppression from #232 for genuinely-done sessions while restoring true positives for unfinished ones.

## Scope

- Applied **byte-identical** across all five copies: `claude` / `codex` / `cursor` / `hermes` + the `pi` extension.
- Contract test updated: asserts the unfinished-work trigger (`MUST write`), the last-message signal, finished-work `none` default, and a **regression guard** that unfinished work is not gated behind the old catastrophe bar. Cross-copy byte-identical check retained (now 5 copies).
- Generation-side prompt only; `resume-brief` parser unchanged. Typecheck clean; contract + pi + resume-brief suites green.

Follow-up to #232 (merged, v0.7.74).

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined next-steps determination logic to use a two-step decision process that better evaluates when session work is complete, defaulting to no next steps when core work is finished while ensuring genuinely unfinished work is properly flagged.

* **Tests**
  * Updated contract tests to verify refined next-steps behavior and session completion detection.


khustup2:
Folded the Next Steps gate rebalance into #233 (it lands on the same SessionStart/resume surface). Closing in favor of that PR.

PR #242, opened 2026-06-07
fix(skillopt): read current skill version by created_at, not the corrupt BIGINT version column

kaghni:
## What

`readCurrentSkillRow` ordered by the BIGINT `version` column; `publishImprovedSkill` does `current.version + 1`. Deeplake's engine returns **corrupted** values for `MAX()`/`ORDER BY` on a BIGINT column in the write-propagation window right after an INSERT — deterministic **12/12** in a tight insert→read loop (phantom values like `281473679796896`≈2^48, present in no row). SkillOpt reads the current version immediately before publishing v+1, so it read garbage and published a garbage version.

## Evidence (full live plugin E2E against the real skillopt org)

A real `claude` session invoked the Skill tool → hooks armed → worker judged → **published `v281473679796898` instead of `v6`**. Root cause isolated: `ORDER BY version DESC`/`MAX(version)` = 12/12 corrupt right after a write; `ORDER BY created_at DESC` = 0/12 (TEXT sorts reliably). Stored data is fine — only BIGINT aggregation/sort during read-after-write is wrong.

## Fix

Order by `created_at DESC, id DESC` (both TEXT) and read the row's own `version`. Validated against the live engine: fixed `readCurrentSkillRow` is **0/8** wrong under tight read-after-write vs 12/12 before. Test asserts the query orders by `created_at` and **not** `version` (negative guard).

## Scope / follow-up

Client-side workaround. The root cause is a Deeplake BIGINT read-after-write bug being fixed in `deeplake-api`. The same `ORDER BY version`/`MAX(version)` pattern exists in `rules/read.ts`, `goal.ts`, `context-renderer.ts`, `deeplake-api.ts` — lower risk (reads aren't post-write-tight), covered by the backend fix.

tsc clean; 16 skillopt tests green.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed a data consistency issue in skill information retrieval that could cause outdated data to be displayed. The system now correctly prioritizes the most recent skill records.

* **Tests**
  * Updated tests to validate the improved skill retrieval logic.

### DIFF
--- a/src/skillify/skill-org-publish.ts
+++ b/src/skillify/skill-org-publish.ts
@@ -63,10 +63,18 @@ export async function readCurrentSkillRow(
     `source_agent, scope, contributors, description, trigger_text, body, version ` +
     `FROM "${sqlIdent(skillsTable)}" ` +
     `WHERE name = '${sqlStr(name)}' AND author = '${sqlStr(author)}' ` +
-    // version DESC, then created_at DESC as a deterministic tie-breaker — if two workers
-    // ever land the same version (cross-machine race), readers resolve the SAME row
-    // instead of an arbitrary one (codex P2).
-    `ORDER BY version DESC, created_at DESC LIMIT 1`,
+    // Order by created_at, NOT version. Deeplake's query engine returns CORRUPTED values
+    // for MAX()/ORDER BY on the BIGINT `version` column when the read lands in the
+    // write-propagation window right after an INSERT — a deterministic 12/12 in a tight
+    // insert→read loop (phantom values like 281473679796896 ≈ 2^48, present in no row).
+    // skillopt reads this immediately before publishing v+1, so `ORDER BY version DESC`
+    // could read a garbage current version and publish a garbage next version (observed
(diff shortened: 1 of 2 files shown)


kaghni:
Closing — **superseded by the pg_deeplake fix.**

This PR worked around Bug #2 (the BIGINT `version` column returning phantom values for `MAX()`/`ORDER BY` in the write-propagation window) by ordering `readCurrentSkillRow` on `created_at` instead. That bug has now been **fixed at the source in pg_deeplake**, so `ORDER BY version DESC` / `MAX(version)` are correct again and this client-side workaround is no longer needed.

Verified against the prod cluster (tight insert→read loop): **0/10 corrupt** (was 12/12 before the fix). The other version readers (rules/goals/skillify-pull) also no longer need any workaround — the backend fix covers them all.

Note: this only fixes Bug #2 (version corruption). Bug #1 (row-visibility lag — just-written rows not immediately in `COUNT`/`SELECT`) is still present (8/10 in the same test); SkillOpt's K=3 judgment window already self-mitigates that, so nothing is blocked.

PR #251, opened 2026-06-09
fix(pre-tool-use): don't intercept incidental memory-path mentions

### LINKED ISSUE #87: pre-tool-use: bash hook intercepts on substring match of memory path
## Severity
Low-medium — UX friction. Hit twice while debugging the embedding_generation branch.

## Where
`src/hooks/pre-tool-use.ts` — `touchesMemory` heuristic + the bash branch path detection in `getShellCommand`.

## Repro
Run a Claude Code subprocess from inside a parent CC session with the memory path string in the prompt argument:

```
claude -p 'Use the deeplake hivemind memory at <home>/.deeplake/memory/...'
```

The OUTER CC's pre-tool-use hook rejects the subprocess invocation with:

```
[RETRY REQUIRED] The command you tried is not available for ~/.deeplake/memory/.
This virtual filesystem only supports bash builtins …
```

The hook intercepted because the path string appeared anywhere in the bash command line — even though it was part of a prompt argument being passed to the inner `claude -p`, not a file argument.

## Root cause
`touchesMemory(cmd)` matches if the path string appears anywhere in the command. It doesn't distinguish between:

- a file argument (`cat <home>/.deeplake/memory/index.md` → should intercept)
- a quoted string inside a prompt or message (`claude -p '... <home>/.deeplake/memory ...'` → should pass through)

## Proposed fix
Tighten the heuristic. The path must appear as a file argument — i.e., as a token immediately following one of the known-bash builtins (`cat`, `ls`, `grep`, `head`, `tail`, `wc`, etc.), not deep inside a quoted string passed to a non-builtin command. Concretely:

- After parsing the bash command into tokens (or splitting on whitespace + quotes), look for the path *only* in the argv positions of the recognized commands.
- If the command is unrecognized (`claude`, `node`, `tsx`, etc.), pass through entirely.

## Acceptance
- `cat <home>/.deeplake/memory/index.md` still intercepted (and routed to virtual FS) ✅
- `claude -p 'use <home>/.deeplake/memory/'` passes through unmodified ✅
- `echo '<home>/.deeplake/memory/'` passes through ✅
- `grep foo <home>/.deeplake/memory/summaries/` still intercepted ✅

Found during embedding_generation branch validation.

Kaustubh1504:
Closes #87

## Problem
The Bash branch ran `touchesMemory(cmd)` as a substring/regex test, so the
memory path triggered interception even when it only appeared as inert text
in a prompt — e.g. `claude -p 'use ~/.deeplake/memory/'` was rewritten to the
`[RETRY REQUIRED]` echo instead of running.

## Fix
Added `bashTouchesMemory()` (in `memory-path-utils.ts`), which tokenizes the
command (`parseBashTokens`) and applies a **narrow carve-out on top of broad
detection**:
- Default: any memory-path reference is a real interaction → intercept. This
  keeps file builtins routed and interpreters/fetchers/substitutions on the
  retry-guidance path (preserves the host-access + path-traversal protections
  the existing tests pin).
- Carve-out: `echo`/`printf`/`claude` stages where the path is inert text pass
  through — UNLESS it's a redirection target (a write, the documented
  `echo '<content>' > '<path>'` path) or smuggles a `$()`/backtick
  substitution, which fall through to be caught.

Swapped `touchesMemory` → `bashTouchesMemory` in the two Bash-command gates in
`pre-tool-use.ts`; bare-path detection for Read/Grep/Glob/Write/Edit is
unchanged.

## Acceptance (#87)
- [x] `cat ~/.deeplake/memory/index.md` — still intercepted
- [x] `claude -p 'use ~/.deeplake/memory/'` — passes through
- [x] `echo '~/.deeplake/memory/'` — passes through
- [x] `grep foo ~/.deeplake/memory/summaries/` — still intercepted

## Why not the argv/command-name approach from the issue
Command-name alone can't satisfy the acceptance: it can't pass `echo '<path>'`
while keeping `rm`/`python3` on a bare memory path routed/guided, since those
are non-reader commands too. Worse, "unrecognized → pass through" would let
`python3 ~/.deeplake/memory/../../../etc/passwd` reach the host shell. The
carve-out approach passes all four cases and keeps the existing security tests
green.

## Tests
Added a `#87` block covering the two passthrough cases plus boundaries (write
redirect, `$()` substitution, quoted reader still intercept). Full existing
suite stays green.

## Out of scope
`parseBashTokens` handles `; | || && \n`, quotes, and `>`/`>>`; it does not
unescape backslashes, fully model `2>`/`&>` fd-redirects, or parse `<(`
(isSafe() rejects `<(`, `$()`, backticks upstream).

## Summary by CodeRabbit

* **New Features**
  * Improved Bash-aware parsing and memory-interaction detection for shell commands, used by pre-tool guidance.

* **Bug Fixes**
  * Avoids false positives for incidental memory-path mentions (e.g., printed, quoted, or heredoc content) while still catching real interactions like redirects, command substitution, and process reads.

* **Tests**
  * Added comprehensive boundary tests for quoting, substitutions, redirections, heredocs, multi-stage commands, and passthrough cases.

* **Chores**
  * CI workflow: tightened condition for posting coverage comments on PRs.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -263,7 +263,11 @@ jobs:
           echo "body-file=/tmp/pr-coverage.md" >> "$GITHUB_OUTPUT"
 
       - name: Post coverage comment on PR
-        if: github.event_name == 'pull_request' && always() && steps.pr-coverage.outputs.body-file != ''
+        # Fork PRs get a read-only GITHUB_TOKEN: the comment API returns
+        # "Resource not accessible by integration", failing the job and
+        # skipping the bundle build/smoke steps below. Skip the comment
+        # there — the coverage gate itself already ran above.
+        if: github.event_name == 'pull_request' && always() && steps.pr-coverage.outputs.body-file != '' && github.event.pull_request.head.repo.full_name == github.repository
         uses: marocchino/sticky-pull-request-comment@v3.0.4
         with:
           header: pr-coverage-report
--- a/src/hooks/memory-path-utils.ts
+++ b/src/hooks/memory-path-utils.ts
@@ -134,7 +134,10 @@ export function parseBashTokens(cmd: string): string[][] {
     const char = cmd[i];
 
     if (escape) { currentToken += char; escape = false; continue; } // prev char was `\`
-    if (char === "\\") { escape = true; currentToken += char; continue; }
+    // Backslash is literal inside single quotes (bash semantics) — treating it
+    // as an escape there would swallow the closing quote and hide a following
+    // `; cmd …` stage inside the quoted token.
+    if (char === "\\" && !inSingle) { escape = true; currentToken += char; continue; }
     if (char === "'" && !inDouble) { inSingle = !inSingle; currentToken += char; continue; } // keep quotes
     if (char === '"' && !inSingle) { inDouble = !inDouble; currentToken += char; continue; }
 
@@ -145,6 +148,13 @@ export function parseBashTokens(cmd: string): string[][] {
       if (char === ">") { // own token, so `>file` (no space) is detectable
         pushToken();
         if (cmd[i + 1] === ">") { currentStage.push(">>"); i++; } else currentStage.push(">");
+        continue;
+      }
+      if (char === "<") { // own token, so `<file` (no space) is detectable;
+        pushToken();      // `<<`/`<<<` (heredoc/herestring) lump into one token
+        let run = "<";    // so only a lone `<` reads a file path
+        while (cmd[i + 1] === "<" && run.length < 3) { run += "<"; i++; }
+        currentStage.push(run);
         continue;
       }
       if (/\s/.test(char)) { pushToken(); continue; }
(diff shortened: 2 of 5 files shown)


efenocchi:
Thanks for this — the approach is right: a narrow carve-out on top of broad detection is the correct call, and your justification for rejecting the "unrecognized command → pass through" idea from #87 is spot on.

While reviewing, we probed the carve-out empirically and found three shapes that the old substring check intercepted but the new code passed through to the host shell. Rather than going back and forth, we pushed a follow-up commit (b19fb20) directly to this branch (maintainer edits are enabled) closing them:

1. **Process substitution** — `echo <(cat <memory-path>)`. The stage-level guard checked `$(` and backticks but not `<(`. The "isSafe() rejects `<(` upstream" note in the description is inverted: `isSafe()` only runs *after* `bashTouchesMemory()` returns true, so the carve-out skipped it entirely. The substitution guard now runs once at the top of `bashTouchesMemory()` on the raw command and includes `<(`.
2. **Input redirection** — `claude -p 'summarize' < <memory-path>` is a genuine memory read, but only `>`/`>>` were tokenized. `<` is now its own token (`<<`/`<<<` are lumped and ignored so heredocs/herestrings are unaffected) and checked as a redirect target.
3. **Backslash inside single quotes** — bash treats `\` as literal there, but the parser treated it as an escape, so `echo 'a\' ; cat <memory-path>` swallowed the `cat` stage into the echo passthrough. The escape branch now skips inside single quotes.

All three are added as regression tests, the boundary tests now also assert the decision shape (per the CodeRabbit note), and the full suite is green (4395 tests). With these in, this looks good to us — thanks again for the contribution and the careful writeup.

PR #254, opened 2026-06-10
fix(capture): exclude `claude -p` from only-cli gate; stop stranding placeholders

efenocchi:
## Problem

Reported by Ivo/Sasun: in the team memory table the **majority of `claude_code` sessions are stuck in the `in progress` state with no summary**. Two independent defects, both reproduced locally:

### A. The `HIVEMIND_CAPTURE_ONLY_CLI` gate let `claude -p` through
The gate used `entrypoint.includes("cli")`. Headless `claude -p` reports `CLAUDE_CODE_ENTRYPOINT="sdk-cli"`, which **contains** `"cli"`, so print-mode sessions slipped past the gate and created capture rows **even with the flag on**. (Confirmed with Sasun's repro script — `BUG: claude -p wrote a placeholder despite HIVEMIND_CAPTURE_ONLY_CLI=true`.)

### B. The wiki worker stranded placeholders under concurrency
`SessionStart` writes an `in progress` placeholder; the detached wiki worker is supposed to replace it with a real summary. But the worker **exited immediately when the sessions table returned zero events**. The capture hooks write events **asynchronously**, and Deeplake reads are eventually-consistent — under concurrency (many short SDK / `claude -p` sessions ending at once) the events lag behind `SessionEnd`, so the worker read 0 events and bailed, leaving the placeholder stuck at `in progress` **forever** (no retry, no cleanup). Reproduced with 10 concurrent Agents-SDK sessions; the "stuck" ones logged `no session events found` while their events were demonstrably present moments later.

(This is also why codex was immune: codex finalizes inline in its always-firing `Stop` hook, with no detached worker racing the async writes.)

## Fix

- **Gate** (`capture-gate.ts`): `entrypoint.includes("cli")` → **`entrypoint === "cli"`**. Only the interactive terminal passes; `sdk-py` / `sdk-ts` / `sdk-cli` (`claude -p`) are excluded.
- **Worker** (`wiki-worker.ts`): on zero events, **retry with backoff** (`HIVEMIND_WIKI_EVENT_RETRIES` / `HIVEMIND_WIKI_EVENT_BACKOFF_MS`, defaults 5 × 1.5s); if events still never arrive, **delete the orphan placeholder** — guarded on `description = 'in progress'` so a concurrent real summary is never clobbered.

No change to the happy path; normal interactive sessions are still captured and summarized.

## Use cases covered

**Gate** — interactive `cli` + flag → captured · `claude -p`/`sdk-cli` + flag → excluded · `sdk-py`/`sdk-ts` + flag → excluded · flag off → all captured · substring-but-not-`cli` → excluded.

**Worker** — events present → finalizes (no regression) · events lag (race) → retry recovers → summary · genuinely 0 events → orphan removed · real summary + 0 events → preserved (guard, no data-loss) · resume of an orphan-deleted session → work → close → correct summary recreated, pointing at the original session jsonl.

## Testing

- **Unit**: `capture-gate.test.ts` updated to exact-match + explicit `sdk-cli` regression; `wiki-worker.test.ts` updated for retry/cleanup + new retry-recovers test. 27/27 green (remaining repo failures are pre-existing, env-only: missing `tree-sitter-*` native deps, CLI TTY mocks — verified identical on clean `origin/main`).
- **E2E vs Deeplake (`*_test` tables)**: gate matrix 4/4 (driving the real `session-start.js`), worker orphan-removal + guard-preserves-real-summary, and a faithful **resume** E2E via the Agents SDK — deleted the orphan, resumed with a real change, and confirmed a correct summary was recreated under the original session path with the right `Source:` pointer.

## Notes / out of scope

- Sasun's comment says `claude -p` reports entrypoint `"cli"`; on claude 2.1.170 I measured **`"sdk-cli"`** (both contain `"cli"`, so the substring bug fires either way). If a runtime ever reports a bare `"cli"` for `-p`, exact-match won't distinguish it and a print-mode signal would be needed — worth a double-check on Sasun's version.
- A **separate, pre-existing** issue (not addressed here): under read-lag, `uploadSummary` can `SELECT path` → miss the not-yet-visible placeholder → `INSERT` a second row, leaving a duplicate `in progress` + summary pair. Follow-up.

## Session Context
[Session transcript](https://gist.github.com/efenocchi/41bdb9c10935e4e6e1a90e164f345bfd)

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced exact CLI-mode detection to prevent misclassifying and capturing non-CLI sessions.
  * Added configurable retry/backoff for session event retrieval to increase reliability.
  * Ensured orphan session placeholders are cleaned up when events never appear to avoid stuck sessions.
* **Tests**
  * Expanded tests to cover strict CLI detection, retry behavior, and orphan-placeholder cleanup.

### DIFF
--- a/src/hooks/shared/capture-gate.ts
+++ b/src/hooks/shared/capture-gate.ts
@@ -1,11 +1,16 @@
 /**
  * Allowlist gate for HIVEMIND_CAPTURE_ONLY_CLI.
  *
- * When the env var is "true", only capture sessions whose
- * CLAUDE_CODE_ENTRYPOINT contains the substring "cli". The Claude Agent SDK
- * (Python / TypeScript) sets the entrypoint to "sdk-py" / "sdk-ts" when it
- * spawns the CLI subprocess, so those sessions fail the check and the hook
- * short-circuits. Interactive terminal sessions keep entrypoint="cli".
+ * When the env var is "true", only capture sessions launched from the
+ * interactive terminal CLI, whose CLAUDE_CODE_ENTRYPOINT is EXACTLY "cli".
+ * Everything else is skipped:
+ *   - "sdk-py" / "sdk-ts"  — Claude Agent SDK (Python / TypeScript)
+ *   - "sdk-cli"            — headless `claude -p` print mode
+ * Matching must be exact equality, NOT a substring test: `claude -p` reports
+ * "sdk-cli", which *contains* "cli", so an `includes("cli")` check would let
+ * print-mode sessions slip through and create stray capture rows even with
+ * the gate on. Interactive terminal sessions are the only ones that report a
+ * bare "cli".
  *
  * Returns true when the gate PASSES (capture should proceed), false when
  * the caller should skip. With the gate disabled (env var unset or != "true")
@@ -20,5 +25,5 @@ export function entrypointPassesOnlyCliGate(
   const onlyCli = env.HIVEMIND_CAPTURE_ONLY_CLI === "true";
   if (!onlyCli) return true;
   const entrypoint = env.CLAUDE_CODE_ENTRYPOINT ?? "";
-  return entrypoint.includes("cli");
+  return entrypoint === "cli";
 }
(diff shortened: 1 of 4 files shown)


efenocchi:
Addressed in 6df6ed80: added a `parseNonNegativeInt` helper so a non-numeric / negative `HIVEMIND_WIKI_EVENT_RETRIES` or `_BACKOFF_MS` falls back to the defaults (5 / 1500ms) instead of becoming `NaN` and silently disabling retries. Added a regression test (`falls back to default retries when the env var is non-numeric`). Good catch — that NaN path would have re-stranded placeholders.

khustup2 (approved):

PR #280, opened 2026-06-20
feat(codex): proactive memory via managed ~/.codex/AGENTS.md block

khustup2:
## What

Gives the **Codex** harness proactive memory use — consult team memory + pull rules/goals before starting — the way claude-code and cursor already have it, but through the channel that actually fits Codex.

## Why AGENTS.md (not an in-context banner)

Codex has **no model-only context channel**: anything the SessionStart hook puts in `additionalContext` is *also* rendered to the user as a `hook context:` cell. So pushing a rules/goals block in-context would be visible noise on every session (the reason it was never added).

`~/.codex/AGENTS.md` solves this: Codex auto-loads it into the model context every session **silently** — no TUI cell. Empirically verified with a sentinel:

```
$ printf '...If asked for the codeword, reply exactly: KIWI_GLOBAL_AGENTS_2026' > ~/.codex/AGENTS.md
$ codex exec --sandbox read-only "What is the codeword?"
codex
KIWI_GLOBAL_AGENTS_2026          # reached the model, never shown as a hook cell
```

## How

- **`src/cli/agents-md.ts`** (new): shared `upsertMarkedBlock` / `stripMarkedBlock` for the marker-fenced hivemind block. `install-pi.ts` refactored to delegate (removes its duplicated copy — net deletion).
- **`install-codex.ts`**: idempotent upsert of the hivemind block into `~/.codex/AGENTS.md` on install, strip on uninstall. The user's own AGENTS.md content outside the `<!-- BEGIN/END hivemind-memory -->` markers is always preserved.
- The block instructs the model to consult `~/.deeplake/memory/` and run `hivemind rules list` / `hivemind goal list --mine` before starting work.
- **`session-start.ts`**: `additionalContext` stays minimal (login + version). No user-visible banner.

## Tests

- Codex AGENTS.md: cold-create (one marker pair, carries the instruction), preserves pre-existing user content, idempotent across 5 installs, uninstall strips block while keeping user content, removes the file when only our block existed.
- session-start guard: asserts no rules/goals block leaks into the user-visible `additionalContext`.
- pi + shared-helper suites stay green after the refactor.
- `tsc --noEmit` clean. (Pre-existing local failures in cli-index/install-consent are an unrelated missing optional dep, `tree-sitter-python` — red on main too.)

## Summary by CodeRabbit

* **New Features**
  * Codex installation now manages a single, idempotent “memory” marker block in `~/.codex/AGENTS.md` (inserted only once) and removes it on uninstall while preserving any existing user content.
  * The Codex session hook continues to emit minimal additional context, preventing the proactive memory/rules-goals banner from appearing.
* **Tests**
  * Added/expanded guard and filesystem tests to verify minimal-context behavior, idempotent insert/remove, duplicate-marker collapse/removal, and `AGENTS.md` cleanup when empty.
  * Improved temp directory fixtures for more secure, reliable filesystem tests.
  * Added unit coverage for marker upsert/strip behavior, including unpaired markers.
* **Maintenance**
  * Consolidated marker-block handling so both Codex and Pi flows use consistent single-block behavior.

### DIFF
--- a/src/cli/agents-md.ts
+++ b/src/cli/agents-md.ts
@@ -1,0 +1,58 @@
+/**
+ * Shared AGENTS.md hivemind-block management.
+ *
+ * AGENTS.md is the global agent-instructions file that several harnesses
+ * auto-load into the model's context every session — SILENTLY, with no
+ * user-visible TUI cell:
+ *   - pi    → `~/.pi/agent/AGENTS.md`
+ *   - Codex → `~/.codex/AGENTS.md`
+ *
+ * We manage only a marker-fenced block and never touch the user's own
+ * content. The marker pair is shared across harnesses (they write to
+ * different files, so there is no collision).
+ */
+
+export const HIVEMIND_BLOCK_START = "<!-- BEGIN hivemind-memory -->";
+export const HIVEMIND_BLOCK_END = "<!-- END hivemind-memory -->";
+
+/**
+ * Insert or replace the hivemind block in `existing`. `block` is the full
+ * marker-fenced block (START … END). Idempotent: a prior block (matched on
+ * the marker pair) is stripped before re-appending, so repeated installs
+ * leave exactly one block. User content outside the markers is preserved.
+ */
+export function upsertMarkedBlock(
+  existing: string | null,
+  block: string,
+  start: string = HIVEMIND_BLOCK_START,
+  end: string = HIVEMIND_BLOCK_END,
+): string {
+  if (!existing) return `${block}\n`;
+  const startIdx = existing.indexOf(start);
+  if (startIdx === -1) return `${existing.trimEnd()}\n\n${block}\n`;
+  const endIdx = existing.indexOf(end, startIdx);
+  // Malformed prior block (no END) — append fresh and let the user clean up.
+  if (endIdx === -1) return `${existing.trimEnd()}\n\n${block}\n`;
+  const before = existing.slice(0, startIdx).trimEnd();
+  const after = existing.slice(endIdx + end.length).replace(/^\n+/, "");
+  const rest = after ? `\n\n${after}` : "";
+  return `${before ? before + "\n\n" : ""}${block}\n${rest}`;
+}
+
+/** Remove the hivemind block from `existing`, preserving surrounding content. */
+export function stripMarkedBlock(
+  existing: string,
+  start: string = HIVEMIND_BLOCK_START,
+  end: string = HIVEMIND_BLOCK_END,
+): string {
+  const startIdx = existing.indexOf(start);
+  if (startIdx === -1) return existing;
+  const endIdx = existing.indexOf(end, startIdx);
+  if (endIdx === -1) return existing;
+  const before = existing.slice(0, startIdx).trimEnd();
+  const after = existing.slice(endIdx + end.length).replace(/^\n+/, "");
+  if (!before && !after) return "";
+  if (!before) return after;
+  if (!after) return `${before}\n`;
+  return `${before}\n\n${after}`;
+}
--- a/src/cli/agents-md.ts
+++ b/src/cli/agents-md.ts
@@ -28,18 +28,16 @@ export function upsertMarkedBlock(
   end: string = HIVEMIND_BLOCK_END,
 ): string {
   if (!existing) return `${block}\n`;
-  const startIdx = existing.indexOf(start);
-  if (startIdx === -1) return `${existing.trimEnd()}\n\n${block}\n`;
-  const endIdx = existing.indexOf(end, startIdx);
-  // Malformed prior block (a BEGIN with no END anywhere) — don't risk
-  // truncating user content; append a fresh block. stripMarkedBlock can still
-  // remove it later (it scans past the malformed marker for well-formed pairs).
-  if (endIdx === -1) return `${existing.trimEnd()}\n\n${block}\n`;
-  // Replace the FIRST block IN PLACE — keeping its position relative to the
-  // user's own notes so their overrides keep their precedence — and strip any
+  const first = findFirstBlock(existing, start, end);
+  // No well-formed block (none at all, or only a stray/unclosed BEGIN) — append
+  // a fresh one. We never pair a stray BEGIN with our block's END, so user text
+  // under a half-written marker is preserved.
+  if (!first) return `${existing.trimEnd()}\n\n${block}\n`;
+  // Replace the first WELL-FORMED block IN PLACE — keeping its position relative
+  // to the user's own notes so their overrides keep precedence — and strip any
   // DUPLICATE blocks from the tail (bad merge / manual paste collapse to one).
-  const before = existing.slice(0, startIdx).trimEnd();
-  const after = stripMarkedBlock(existing.slice(endIdx + end.length), start, end)
+  const before = existing.slice(0, first.startIdx).trimEnd();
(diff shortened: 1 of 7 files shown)


khustup2:
@coderabbitai review

khustup2:
## Review responses

**CodeRabbit nitpicks** — both addressed:
- *Import markers from source* (`cli-install-codex-fs.test.ts`) → fixed in `ccaa7325`: the test now imports `HIVEMIND_BLOCK_START/END` from `src/cli/agents-md.ts` instead of duplicating the literals.
- *Use specific-value assertions* (`codex-session-start-hook.test.ts:133-134`) → fixed in `ccaa7325`: replaced the two `not.toContain` negatives with one anchored full-pattern match on the exact logged-in `additionalContext`.
- *Docstring coverage* → `007fa194`: `writeCodexAgentsBlock` / `removeCodexAgentsBlock` now have `/** */` docstrings.

**Codex [P2] "Emit Windows-safe AGENTS guidance"** — respectfully declining, with rationale:

1. **The "bash builtins only" constraint is a property of the deeplake-memory VFS, not the host shell.** The memory mount enforces *only* `cat/ls/grep/jq/...` at the interceptor layer on every OS (it rejects `python`/`node`/`curl`/`rg` regardless of platform). So the guidance is *correct* on Windows; gating it on `process.platform` would hand Windows users *inaccurate* guidance about the VFS contract.
2. **Consistency.** Every existing harness memory block (pi `~/.pi/agent/AGENTS.md`, claude-code, cursor — claude-code already runs on Windows) uses the identical POSIX language, unguarded. Platform-gating Codex alone would diverge from the established cross-harness convention and falsely imply the others are "fixed."
3. If Windows-specific memory guidance is genuinely wanted, it's a **cross-harness** change to all blocks at once — out of scope for this Codex-only PR.

Re-running codex + re-triggering coderabbit on the updated diff.

khustup2:
@coderabbitai review

khustup2:
@coderabbitai review

khustup2:
## Review responses (round 2)

**CodeQL — 4 high-severity alerts** → fixed in `ab7d3479`:
- `js/file-system-race` (×2, install-codex.ts): replaced `existsSync()`-then-read with a `readFileOrNull()` helper (direct read, catch ENOENT) — removes the TOCTOU window.
- `js/insecure-temporary-file` (×2, test): switched the temp root to `mkdtempSync(join(tmpdir(), "hm-codex-"))`, matching the prior repo fix `2ffc08f2`.

**Codex [P2] ×2 — "remove/strip ALL hivemind blocks"** → fixed in `715807f6`:
- `upsertMarkedBlock` now strips *every* existing block before re-appending one (idempotent even for files with duplicate marker pairs from a bad merge/manual paste).
- `stripMarkedBlock` now loops until all pairs are removed, while still leaving a malformed `BEGIN`-without-`END` untouched (never truncates user data).
- Benefits pi + codex (shared helper); added duplicate-block regression tests for both.

Re-running codex + CodeRabbit on `715807f6`.

khustup2:
@coderabbitai review

khustup2:
CodeRabbit nitpick (exact-output assertions in the duplicate-block tests) → fixed in 051f9771. CodeQL is green (all 4 high alerts cleared) and codex is clean. Re-triggering CodeRabbit for a final pass.

khustup2:
@coderabbitai review

khustup2:
## Review responses (round 3)

**Codex [P2] "Preserve block position on reinstall"** → fixed in `fdef754d`. The helper now replaces the **first** block *in place* (keeping its position relative to the user's own notes, so their overrides keep precedence) while still stripping duplicate blocks from the tail. Added an in-place regression test asserting the block stays above later user notes.

**Codex [P3] "Malformed-BEGIN orphan"** → not a live bug, here's the trace. `stripMarkedBlock` does **not** stop-and-return-unchanged when an installed block follows a malformed marker: `indexOf(END, beginIdx)` finds the installed block's `END`, so the pair is matched and the block is removed (uninstall is not orphaned). The loop only `return`s early once the *only* remaining `BEGIN` has no `END` at all — i.e. the user's own half-written marker, which we deliberately leave untouched rather than truncate their content. Existing "malformed → unchanged" tests cover that contract.

Re-running codex + CodeRabbit on `fdef754d`.

khustup2:
@coderabbitai review

khustup2:
## Re: Codex [P2] "Continue past malformed markers"

This is a **false positive** — I added a round-trip regression test (`fdef754d`+1) reproducing the exact scenario, and the block **is** removed:

```ts
const userFile = `# Notes\n${BEGIN}\nuser half-wrote this, no end\n`; // stray malformed BEGIN
const installed = upsertHivemindBlock(userFile);   // appends a well-formed block
const uninstalled = stripHivemindBlock(installed);
expect(uninstalled).not.toContain(END);            // ✅ passes — block removed, not orphaned
```

Structural reason `stripMarkedBlock` never orphans our block: it only `return`s early when a `BEGIN` has **no `END` after it**. But the appended block always supplies an `END` after any earlier (malformed) `BEGIN`, so `indexOf(END, beginIdx)` finds it and the loop removes *through* that range — taking our block with it. For our block to survive, it would have to sit after a `BEGIN` that has no following `END`, which is impossible once our block (which contains an `END`) exists.

The only thing `strip` intentionally leaves is a lone user-authored `BEGIN` with no `END` anywhere and no managed block present — i.e. nothing of ours to remove. Considering this finding refuted with the passing test as evidence.

khustup2:
@coderabbitai review

khustup2:
## Re: Codex [P2] "Avoid pairing a stray BEGIN with our appended END"

**You're right, and my previous refutation was wrong** — I verified "block removed" but missed that pairing the stray `BEGIN` with our block's `END` deletes the user's lines *in between*. That's real data loss. Fixed in `4e8214c6`.

`stripMarkedBlock` now only pairs a `BEGIN` with an `END` when **no other `BEGIN` appears before that `END`**. A stray/unclosed `BEGIN` is skipped instead of swallowing a later block's `END`, so user content between the stray marker and our block is preserved. Updated the round-trip test to assert exactly that:

```ts
const userFile = `# Notes\n${BEGIN}\nuser half-wrote this, no end\n`;
const uninstalled = stripHivemindBlock(upsertHivemindBlock(userFile));
expect(uninstalled).not.toContain(END);                       // our block removed
expect(uninstalled).toContain("user half-wrote this, no end"); // ✅ user content preserved
```

Thanks for holding the line on this one.

khustup2:
@coderabbitai review

khustup2:
Codex [P2] reinstall data-loss → fixed in 7a8d928f. Extracted `findFirstBlock()` (first BEGIN whose END has no intervening BEGIN — stray markers skipped) and routed BOTH `upsertMarkedBlock` and `stripMarkedBlock` through it, so neither install nor uninstall swallows user text under a half-written marker. Added a reinstall regression test asserting the user's text is preserved.

khustup2:
@coderabbitai review

khustup2:
CodeRabbit nitpick (exact-output in the data-loss round-trip test) → fixed in 9c029b93: now asserts `uninstalled === userFile` (uninstall restores the user's file byte-for-byte). All CI green, CodeQL clean (0 alerts), codex clean. Re-triggering CodeRabbit for a final pass.

khustup2:
@coderabbitai review

efenocchi (approved):

PR #287, opened 2026-06-24
fix: make proactive recall reliably deliver value (5 root-cause fixes)

khustup2:
## Make proactive recall actually deliver value (5 fixes)

Built an isolated-Docker A/B harness that measures whether Hivemind's
cross-session memory changes a second session's answer. It surfaced that recall
*rarely fires in practice*, and production confirms it: of **4,732** real
session summaries, only **11.5%** are recall-eligible — **56% stuck
`description='in progress'`** (never finalized) and only **25% embedded**. This
PR fixes the five root causes. After them, the A/B value-win rate goes **1/9 →
5/9**, "recall fired but answered wrong" goes **80% → 0%**, and recall firing on
a cross-person demo goes **1/3 → 3/3**.

Stacked on `feat/proactive-recall` (the diff is exactly these 5 commits).

### The fixes
1. **recall fidelity** (`recall-format/gate/query.ts`, `spawn-wiki-worker.ts`) —
   inject a verbatim `## Key Facts` excerpt (not just the gist `description`);
   preserve exact identifiers in the summary prompt; threshold 0.55→0.50
   (env-tunable). Kills the "recall fired but dropped the exact fact" failure.
2. **finalize-wins** (`upload-summary.ts`) — a placeholder/empty write can no
   longer downgrade a finalized summary back to a stub.
3. **embedding coverage** (`embed-summary.ts`, `wiki-worker.ts`) — warm the embed
   daemon + retry on finalize so a cold daemon no longer strands
   `summary_embedding = NULL` forever (there was no backfill).
4. **recall warm-daemon** (`recall.ts`) — warm the daemon + retry the *query*
   embed so the **first prompt of a session** doesn't miss semantic recall
   (cold-daemon race); budget 1000→1500ms, still under the 2s hook timeout.
5. **summary revert race** (`shared/placeholder-summary.ts` + 4 session-start
   variants) — **the root blocker**: `createPlaceholder` did SELECT-then-INSERT
   and the `memory` table has no unique constraint on `path`, so a stale read on
   a second SessionStart inserted a *duplicate* `in progress` stub that shadowed
   the finalized row (unordered `LIMIT 1` reads). Replaced with an atomic
   `INSERT … SELECT … WHERE NOT EXISTS`. This is the mechanism behind the 56%
   stuck-in-progress production stat.

### Tests
Each fix is red→green with unit tests (vitest): finalize-wins, embed-summary
warmup/retry, recall fidelity/threshold, placeholder revert-race (incl. the
stale-read duplicate case), and an updated EmbedClient mock. Target suites green;
remaining whole-suite failures are pre-existing env issues (tree-sitter native
addons, network smoke tests) unrelated to these changes.

### Validation
Measured against the live backend in isolated containers (one per `claude`
session; continuity only through Deeplake). Reproducible harness + before/after
demo cards available separately.

### DIFF
--- a/src/embeddings/embed-summary.ts
+++ b/src/embeddings/embed-summary.ts
@@ -1,0 +1,75 @@
+// Robust summary embedding for the finalize (wiki-worker) path.
+//
+// Background: ~75% of production summary rows have a NULL summary_embedding,
+// so proactive recall silently degrades to weak lexical matching for most of
+// the corpus. A large share of those NULLs are NOT "embeddings disabled" — they
+// are ENABLED users whose embed daemon was cold at the one moment finalize ran.
+//
+// EmbedClient.embed() is built for the latency-critical capture hook: on a cold
+// daemon it returns null IMMEDIATELY while spawning the daemon in the
+// background ("fire-and-forget on miss"). The wiki-worker is a detached
+// background process, NOT latency-critical — so paying a one-time warmup +
+// retry here turns the common cold-start NULL into a real embedding, with no
+// effect on the hot capture path and no change to the opt-in semantics
+// (callers still gate on embeddingsDisabled() before invoking this).
+//
+// Returns the embedding vector, or null when the daemon truly cannot be
+// reached / embeddings are unavailable. Never throws — finalize must always
+// proceed and write the summary, with NULL embedding as the graceful fallback.
+
+import { EmbedClient } from "./client.js";
+import type { EmbedKind } from "./protocol.js";
+
+/** Subset of EmbedClient the helper depends on — lets tests inject a fake. */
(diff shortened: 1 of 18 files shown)


khustup2:
Superseded by #288. The root cause of unreliable recall was a pg-deeplake read-consistency bug (in-place UPDATE'd summary rows reading back stale), fixed separately on the backend. With that fixed, #288 carries the minimal, non-embedding improvements that help every path including the no-embeddings (lexical) default — finalize-wins, atomic placeholder, and recall content fidelity. The embedding-specific changes here (recall warm-daemon, embedding-coverage warm+retry, model-swap) are deferred to separate branches pending a decision to prioritize the semantic path. Closing this in favor of #288.

PR #288, opened 2026-06-25
Recall quality + summary integrity (minimal, non-embedding)

khustup2:
## Recall quality + summary integrity (minimal, non-embedding)

The core reason recall was unreliable turned out to be a **pg-deeplake read-
consistency bug** (an in-place `UPDATE`'d summary row read back its stale
pre-UPDATE value non-deterministically) — fixed separately on the backend. With
that fixed, this PR carries only the **plugin-side improvements that help every
path, including the no-embeddings (lexical) default**:

1. **finalize-wins** (`upload-summary.ts`) — a stale/empty write can never
   downgrade a finalized summary back to a stub.
2. **atomic placeholder** (`shared/placeholder-summary.ts` + the codex/cursor/
   hermes session-start variants) — `createPlaceholder` is now a single
   `INSERT … SELECT … WHERE NOT EXISTS`, so a stale read can't wedge a duplicate
   `in progress` stub that shadows the finalized row (defensive hardening; the
   backend fix removed the original trigger).
3. **recall content fidelity** (`shared/recall-format.ts`, `recall-query.ts`,
   `spawn-wiki-worker.ts`) — the wiki prompt preserves verbatim
   identifiers/facts; recall injects a real **Key Facts excerpt** (not just the
   one-line gist) and searches/surfaces the `summary` field. This improves both
   lexical matching and the quality of what the agent receives — independent of
   embeddings.

Also makes the cosine threshold operator-tunable (`HIVEMIND_RECALL_THRESHOLD`)
**without changing the 0.55 default** (a looser 0.50 is deferred pending
validation — it's semantic-only and didn't reliably help in measurement).

**Deliberately NOT included** (deferred — embedding-only, and the lexical path is
the default): the embed-daemon warmup/retry on recall, the embedding-coverage
warm+retry on finalize, and the model-swap (bge/MiniLM) enabler. Those live on
separate branches and can land if/when the semantic path is prioritized.

Supersedes #287 (which bundled the embedding-specific fixes). Stacked on
`feat/proactive-recall`; tests green (132 across the touched suites; red→green
for finalize-wins and the placeholder revert-race).

### DIFF
--- a/embeddings/embed-daemon.js
+++ b/embeddings/embed-daemon.js
@@ -67,10 +67,14 @@ var NomicEmbedder = class {
   repo;
   dtype;
   dims;
+  docPrefix;
+  queryPrefix;
   constructor(opts = {}) {
     this.repo = opts.repo ?? DEFAULT_MODEL_REPO;
     this.dtype = opts.dtype ?? DEFAULT_DTYPE;
     this.dims = opts.dims ?? DEFAULT_DIMS;
+    this.docPrefix = opts.docPrefix ?? DOC_PREFIX;
+    this.queryPrefix = opts.queryPrefix ?? QUERY_PREFIX;
   }
   async load() {
     if (this.pipeline)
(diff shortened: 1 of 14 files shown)


khustup2:
Closing — #287 already merged, so the non-embedding fixes are in feat/proactive-recall. The merged embedding code is dormant for the lexical default (opt-in). Reverting just the cosine threshold default to 0.55 directly on feat/proactive-recall (no PR needed); the looser 0.50 stays available via HIVEMIND_RECALL_THRESHOLD.

PR #298, opened 2026-07-01
fix: stop wiki summary worker crash on long sessions

efenocchi:
## Problem

The Stop hook's background wiki-summary worker crashed on long sessions (reported on a ~4480-event Codex session): `ENOBUFS` (the agent's stdout exceeded Node's 1 MB default `execFileSync` buffer) and `ETIMEDOUT` (120s) because the worker re-fed the **entire** session to the summarizer on every run.

Root cause: the offset that should bound each run to only the new rows was never applied — no `rows.slice()` existed, so the full session was always written to the JSONL; and the offset value was parsed by regex from the LLM-generated summary text (fragile) instead of the reliable sidecar counter. When the summary failed, `finalizeSummary` never ran, so the offset stayed at 0 and every run reprocessed everything.

## Fix (all 5 workers: claude-code, codex, cursor, hermes, pi)

- `maxBuffer: 64 MB` on the summarizer `execFileSync` → eliminates ENOBUFS (the summary is written to a file, not read from stdout).
- Read the offset from the **sidecar** (`readState().lastSummaryCount`), with the regex on the stored summary kept only as a cross-machine fallback.
- Actually slice `rows.slice(prevOffset)` → only new rows reach the agent; the run is skipped entirely when nothing is new.
- New shared helper `src/hooks/wiki-offset.ts`:
  - `stampOffset` — the worker writes the offset into the summary itself, removing the dependency on the LLM echoing a bookkeeping line.
  - `capLinesByBytes` — 4 MB tail cap safety net (keeps the newest rows).
- Added the `execSucceeded` guard to cursor/hermes/pi/claude-code (codex already had it) so a failed run on a **resumed** session never advances the offset past rows that were never summarized.

OpenClaw is N/A (pure HTTP/WebSocket gateway — does not run the wiki-worker daemon). MCP/Cowork reuses the base worker.

## Known tradeoff

When a single increment exceeds 4 MB (degenerate case: first summary of an already-huge backlog with offset 0, or a lone giant row) the byte cap keeps the newest rows and permanently skips the oldest ones (logged, not silent). Intentional — documented in the `capLinesByBytes` doc comment. In normal incremental operation increments are tiny and nothing is ever dropped.

## Tests

`tsc` clean; 77 worker/helper tests pass. New coverage: offset slicing feeds only new rows, byte-cap keeps the tail and reports drops, no-new-rows skip, `maxBuffer` presence, and the failure guard that prevents offset advance (data loss) on a failed resumed run.

## Summary by CodeRabbit

* **New Features**
  * Wiki summarization resumption now uses a persisted per-session progress counter as the authoritative resume point, processing only newly added events.
  * Added stricter per-run JSONL byte limiting with byte-safe truncation and dropped-line tracking.
* **Bug Fixes**
  * Prevents offset advancement, uploads, and finalization when the summarizer run fails or produces unchanged/empty output.
  * Skips summarization when there are no new events to process.
  * Reduces subprocess failures by increasing stdout buffering capacity.
* **Tests**
  * Expanded resume/offset, JSONL capping, buffering, and failure-mode coverage across workers, plus new unit tests for offset stamping and capping.

### DIFF
--- a/src/hooks/codex/wiki-worker.ts
+++ b/src/hooks/codex/wiki-worker.ts
@@ -176,7 +176,14 @@ async function main(): Promise<void> {
         writeFileSync(tmpSummary, existing);
         hasExistingSummary = true;
       }
-    } catch { /* no existing summary */ }
+    } catch (e: any) {
+      // A genuine lookup failure (query() throws only after its own retries) is
+      // NOT the same as "no summary". Treating it as absent would slice to the
+      // newest rows and overwrite the canonical summary with a base-less one, so
+      // bail and retry on the next run instead.
+      wlog(`existing summary lookup failed: ${e.message}; skipping to avoid overwriting the base summary`);
+      return;
+    }
     // The offset only means something if we actually loaded the summary it
     // refers to. If the summary row is gone (or the read failed), slicing by a
     // stale sidecar count would drop old rows with no base summary to extend,
--- a/src/hooks/codex/wiki-worker.ts
+++ b/src/hooks/codex/wiki-worker.ts
@@ -190,7 +190,7 @@ async function main(): Promise<void> {
     const newLines = newRows.map(r => typeof r.message === "string" ? r.message : JSON.stringify(r.message));
     const { kept, dropped } = capLinesByBytes(newLines, WIKI_JSONL_MAX_BYTES);
     if (dropped > 0) {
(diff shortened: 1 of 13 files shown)


efenocchi:
@coderabbitai review

khustup2 (approved):

PR #302, opened 2026-07-10
feat: per-directory .hivemind config (route org/workspace, opt out of capture)

khustup2:
## What

Adds an in-tree **`.hivemind`** file that lets a directory tree override the global Hivemind identity — the two things people kept asking for:

1. **Route** this tree's captured traces to a specific org/workspace.
2. **Opt out** of capture entirely (`collect: false`).

```json
// route this repo's traces to a client org/workspace
{ "orgId": "acme-corp", "workspaceId": "client-work" }

// never collect traces from this folder
{ "collect": false }
```

It travels with the repo like `.editorconfig`; a gitignored **`.hivemind.local`** handles personal overrides (the `.env.local` convention). A copy-ready `.hivemind.example` is included.

## How it resolves

- Walk **up** from `cwd`, **nearest-wins** (the `.git`/`.gitconfig` model, *not* `.gitignore`-style merge). `.hivemind.local` beats `.hivemind` in the same dir.
- **Precedence is the conventional `env > file > login`** — an explicit `HIVEMIND_ORG_ID` / `HIVEMIND_WORKSPACE_ID` locks that field.

## Safety by disclosure, not ceremony

No trust/allow commands. Two properties make routing safe:

- **`.hivemind` never carries a token** — auth stays in `~/.deeplake/credentials.json`, so a routing override can only ever target orgs your login *already* authorizes (the API rejects anything else). Worst case is misfiling into another of your own orgs, not leaking to a stranger.
- **Every session-start banner prints the effective identity** — `org: … (workspace: …) · routed by ./.hivemind`, or `capture is disabled for this directory` — so a redirect is never silent.

## Scope

Wired across **all four agents** (claude/codex/cursor/hermes): `capture`, `session-end`, `session-start` (placeholder + banner), and `graph-push` (new `skipped-collect-disabled` outcome). `collect:false` *skips* the placeholder write rather than adding new machinery.

## Tests

- **Resolver units** (`tests/shared/dir-config.test.ts`): walk-up, nearest-wins, `.local` precedence, malformed-file skip, whitelisting, `collect:false`, overlay, env precedence.
- **Hook wiring** (`capture` + `session-start`): a real `.hivemind` under the mocked `cwd` proves the primary write path writes **nothing** on `collect:false` and builds the API against the **routed** org/workspace, with `env > file` verified end-to-end.

Full suite green except pre-existing env-only failures (`tree-sitter-python` not installed → graph/\*; flush-memory network timeout) — both untouched here. Bundles are gitignored (built at pack time), so the diff is source + tests + README.

## Summary by CodeRabbit

* **New Features**
  * Added per-directory `.hivemind` / `.hivemind.local` config to route org/workspace identity and optionally disable capture (`collect:false`).
  * Cloud sync now skips when the resolved directory disables collection.
* **Bug Fixes**
  * Capture, session setup, and session summary flows now consistently honor directory-level opt-outs and routing, including cwd-aware resolution.
* **Documentation**
  * Documented `.hivemind` format, precedence/overlay rules, and interaction with global settings.
* **Tests**
  * Added/updated filesystem-backed tests for parsing/resolution and capture gating.
* **Chores**
  * Added `.hivemind.local` to `.gitignore` and included an example config file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ tmp/
 *.d.ts.map
 .env
 .env.*
+# Personal per-directory Hivemind override (routing / opt-out). The committed
+# `.hivemind` is shared; `.hivemind.local` is yours alone — never commit it.
+.hivemind.local
 coverage/
 jscpd-report/
 bench/
--- a/.hivemind.example
+++ b/.hivemind.example
@@ -1,0 +1,5 @@
+{
+  "orgId": "your-org-id-or-name",
+  "workspaceId": "your-workspace",
+  "collect": true
+}
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ Disable capture entirely:
 HIVEMIND_CAPTURE=false claude
 ```
 
-Disable capture for a specific directory tree (persistent, travels with the repo) — drop a `.hivemind` file with `{ "collect": false }`. See [Per-directory config](#per-directory-config-hivemind).
+Disable capture for a specific directory tree (persistent, travels with the repo) by dropping a `.hivemind` file with `{ "collect": false }`. See [Per-directory config](#per-directory-config-hivemind).
 
 Enable debug logging:
 
@@ -327,7 +327,7 @@ This plugin captures session activity and stores it in your Deeplake workspace:
 
 ## Per-directory config (`.hivemind`)
 
-The variables above set **one global identity** for the whole machine. A `.hivemind` file lets a specific directory tree override that — either to **route** its traces to a different org/workspace, or to **opt out** of capture entirely.
+The variables above set **one global identity** for the whole machine. A `.hivemind` file lets a specific directory tree override that: either **route** its traces to a different org/workspace, or **opt out** of capture entirely.
 
 Drop a `.hivemind` JSON file at the root of the tree you want to configure:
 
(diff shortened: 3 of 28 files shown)


efenocchi:
@khustup2 I'd check Pi and openclaw as well

khustup2:
@efenocchi good call. Wired **pi** (`harnesses/pi/extension-source/hivemind.ts`) — it's a local agent with a real cwd, so it now honors `.hivemind` on every capture path (input / tool_result / message_end / session_shutdown) plus session_start gating + disclosure. Since pi ships as raw .ts with no shared imports, it carries a compact inline mirror of `dir-config.ts` (nearest-wins, `.local` beats committed, env > file), covered by source-text tests.

**openclaw** I left out on purpose: it's a channel gateway with no per-project filesystem cwd (`cwd: homedir()` sentinel, see index.ts), and `.hivemind` is a per-directory concept — there's no project dir to place one in. Routing/opt-out for openclaw would go through global creds/env, not `.hivemind`. Shout if you'd rather I handle it differently.

efenocchi (approved):

khustup2:
@efenocchi fixed 🙏

**openclaw CI failure** ("Audit openclaw bundle" step): my `dir-config.ts` used `env = process.env` as a default param — an *alias* that reached the openclaw bundle via `deeplake-push → dir-config` as a bare `process.env` substring, tripping ClawHub's env-harvesting scan. Switched to literal `process.env.HIVEMIND_ORG_ID`/`WORKSPACE_ID` reads (which the openclaw esbuild `define` stubs to `undefined`); tests inject via an `envOverride` arg. Verified locally: `audit:openclaw --criticals-only` → 0 critical.

**CodeRabbit minors:**
- `dir-gate.ts`: log the resolved cwd (`cwd || process.cwd()`) instead of `?` on empty cwd.
- `pi`: it referenced `HIVEMIND_ORG_ID` for the lock but never folded env into creds — now folds env (env > file > login), so pi matches the other agents' precedence and the lock reflects an applied value.
- The cursor "display resolved identity" comment was already addressed in the earlier review pass (uses `effConfig = dirRes?.config ?? baseConfig`).

Green locally (typecheck + touched suites).

PR #309, opened 2026-07-12
fix(recall): semantic-only proactive recall, drop ILIKE full-scan fallback

khustup2:
## Problem

Proactive recall (`UserPromptSubmit` hook) fell back to a **lexical ILIKE** search when semantic embeddings were unavailable or missed the threshold. That fallback (`recallTopHitLexical`) issued:

```sql
... WHERE path LIKE '/summaries/%' AND (summary||description) ILIKE '%kw%' OR ...
```

A leading-`%` `ILIKE` can't use any index, so pg-deeplake runs it as a **full-table scan over the summary/message text columns** — measured at **~37s** on a large org (`coding.sessions` ≈ 923k rows). Worse, the recall client abandons the query at the 1500ms budget, but the backend keeps executing to completion, so **the server pays the full multi-second cost even though the result is thrown away**.

## Fix

Remove the lexical fallback entirely. Recall is now **semantic (cosine) only**:

- If embeddings are unavailable, yield no vector, or the top hit misses the threshold → **skip** (emit nothing).
- No ILIKE query is ever issued from the proactive path, so it can't trigger the backend full-scan.

The semantic path, the 1500ms recall budget, and failure-isolation are unchanged.

## Changes
- `recall.ts` — `findHit()` semantic-only; drop `extractKeywords`/`MIN_LEXICAL_OVERLAP`/`recallTopHitLexical` usage; simplify `hitPasses` + log bar
- `recall-query.ts` — delete `recallTopHitLexical` + now-unused `sqlLike` import
- `recall-format.ts` — narrow `RecallHit.mode` to `"semantic"`
- `tests/shared/recall.test.ts` — drop the lexical-path suite

Net **−78 lines**. `tsc --noEmit` clean; all **61** recall tests pass.

## Notes
- `extractKeywords` / `MIN_LEXICAL_OVERLAP` in `recall-gate.ts` are now unused in `src` (kept to keep this focused; can delete in a follow-up).
- The **reactive** grep path (`virtual-table-query.ts`, the memory+sessions UNION ILIKE) is the other source of the same full-scan and is **not** touched here — separate follow-up.

## Summary by CodeRabbit

* **Changes**
  * Proactive recall now uses semantic matching exclusively (no keyword/lexical fallback).
  * Recall results now consistently report semantic similarity scoring and semantic-only mode.
* **Bug Fixes**
  * Prevented unavailable or below-threshold semantic searches from triggering broad keyword scans.
* **Tests**
  * Updated recall test coverage to validate semantic-only behavior, including scenarios where embeddings are disabled or unavailable.

### DIFF
--- a/src/hooks/recall.ts
+++ b/src/hooks/recall.ts
@@ -102,9 +102,10 @@ function emit(additionalContext: string): void {
 }
 
 /**
- * Find the top hit: semantic when embeddings yield a vector, else lexical.
- * Also falls back to lexical when semantic finds nothing (e.g. summaries not
- * embedded yet). Bounded by withDeadline in main.
+ * Find the top hit: SEMANTIC-ONLY. Embeds the prompt and takes the top cosine
+ * hit that clears the threshold. If embeddings are unavailable, the prompt
+ * yields no vector, or nothing clears the bar, recall is skipped — there is
+ * deliberately no lexical (ILIKE) fallback. Bounded by withDeadline in main.
  */
 async function findHit(
   input: RecallInput,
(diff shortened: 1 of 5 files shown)


khustup2:
@coderabbitai the stale lexical-fallback comments (recall.ts:146-147 self-heal block + the `findHit` docstring) are fixed in e98aa533 — they now say recall is skipped entirely when the symlink self-heal / embeddings are unavailable, with no lexical degradation path. Same commit rewrites tests/claude-code/recall-hook.test.ts to assert semantic-only behavior (no ILIKE query on empty/dead-daemon paths). Please re-review.

khustup2:
@coderabbitai review

khustup2:
Codex review (local, `codex review --base origin/main`) on e98aa533 raised two P1s, both objecting to the PR's intended design rather than a bug:

1. "Preserve proactive recall when embeddings unavailable" and 2. "Keep un-embedded historical summaries recallable" — both amount to *keep the ILIKE lexical fallback*. That fallback is exactly what this PR removes on purpose: it forced unindexed full-table scans on pg-deeplake (seconds/prompt). The behavior is deliberate and documented in the `findHit` docstring — semantic-only, skip when semantic is unavailable, "better to return nothing."

Not re-introducing the fallback (that's the whole point of the PR). Flagging the rollout implication to the owner: embeddings are opt-in, so recall goes dark for users without `hivemind embeddings install` and for un-embedded historical rows until backfilled — an accepted tradeoff vs. the scan latency.

PR #317, opened 2026-07-17
fix(skillify): route push/pull through .hivemind, not the global workspace

khustup2:
## What DevOps hit

Artyom set up a directory with a `.hivemind` config, but saving a skill from it wrote to the **global** workspace, not the routed one. His words in #devops:

> saved a local skill to deeplake, skill got appeared in default workspace

That test happens to exercise the one writer that never got routed.

## Root cause

`src/commands/skillify.ts` read `loadConfig()` raw at the push and pull sites and passed the global `workspaceId` straight to `DeeplakeApi` — no `resolveDirConfig`. Session capture, session-end summaries, and graph push all route through `.hivemind` already; **skillify was a straggler nobody threaded.** The `DeeplakeApi` constructor puts the workspace in the URL path (`/workspaces/{workspaceId}/tables/query`), so that argument decides where the row lands — and it was always the global one.

To be clear about scope: **regular session capture routes correctly** (traced end-to-end: `resolveCaptureConfig` → `workspaceId` → URL path → write). Artyom's complaint was phrased about sessions but the only thing he tested was skillify, which is the single broken path. This PR fixes that path.

## The fix

Resolve the nearest `.hivemind` at both the push and pull sites, so skills write to — and read from — the directory's routed workspace, symmetric with where that tree's traces are captured.

- `orgId`/`workspaceId` are identity → apply to reads and writes.
- `collect:false` gates capture only, so a routed + opted-out directory still pushes/pulls to its workspace.
- **unpull is untouched** — it's a local-filesystem operation that never queries a workspace (only reads `userName` for `--not-mine`).

## Test

`tests/claude-code/skillify-dir-routing.test.ts` runs `resolveDirConfig` for real against the process cwd and asserts which workspace `DeeplakeApi` was constructed against — the URL path that decides where the write goes. Covers push + pull, ancestor inheritance, `collect:false` still routing, and the no-`.hivemind` global fallback. **Verified to fail against the pre-fix code** (the 4 routing assertions fail; the 2 global-fallback cases pass in both, as they should).

## Verification

- Typecheck clean.
- All skillify suites green; full suite has **no new failures** (the 14 pre-existing failures are the `tree-sitter-python`-missing graph suites + the `flush-memory` network timeout, all unrelated).
- No coverage threshold on `skillify.ts`; coverage run clean.

## Still open (not in this PR)

The rules/goals CLI writers use the same `loadConfig()`-raw pattern and will behave identically — nobody's tested them yet, but they're the next straggler. Happy to fold them into a follow-up.

https://claude.ai/code/session_017SDmiDuUn7Wqhi7qdn1p1f

## Summary by CodeRabbit

* **Bug Fixes**
  * `skillify push` and `skillify pull` now honor the nearest `.hivemind` workspace configuration.
  * Commands correctly inherit workspace settings from parent directories.
  * Added fallback to the global workspace when no directory-specific configuration is available.
  * Improved handling of `collect:false` while preserving the selected workspace.

### DIFF
--- a/src/commands/skillify.ts
+++ b/src/commands/skillify.ts
@@ -30,6 +30,7 @@ import { runPull, type PullSummary } from "../skillify/pull.js";
 import { runPush } from "../skillify/push.js";
 import { runUnpull } from "../skillify/unpull.js";
 import { loadConfig } from "../config.js";
+import { resolveDirConfig } from "../dir-config.js";
 import { DeeplakeApi } from "../deeplake-api.js";
 import { runMineLocal } from "./mine-local.js";
 import { renderSubcommandUsageBlock } from "../cli/skillify-spec.js";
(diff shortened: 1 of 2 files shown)


khustup2:
Superseded by #318, which folds this skillify fix into a single-source-of-truth refactor (one loadRoutedConfig entry point + a guard that prevents the whole class of unrouted-writer bugs).

Issue #331, opened 2026-07-29
wiki-worker spawns a headless `claude -p` per capture event: flashing console windows on Windows, ~5x process churn, continuous background API calls


jzferrell26:
# Bug report: `wiki-worker` spawns a headless Claude Code process per capture event, flashing console windows on Windows

**Component:** `@deeplake/hivemind` Claude Code plugin
**Affected versions:** 0.7.131 and 0.7.142 (spawn logic identical in both)
**Platform:** Windows 11 Home 26200 (behavior is Windows-specific in its visible symptom; the underlying spawn cost is cross-platform)
**Severity:** Medium. Not data-losing, but continuously disruptive, and it consumes CPU and model API calls in the background.

---

## Summary

While a Claude Code session is doing any tool work, `bundle/capture.js` launches `bundle/wiki-worker.js`, and that worker launches a **complete headless Claude Code process** (`claude -p --model <model> --permission-mode bypassPermissions`) on a repeating basis.

On Windows, each of those child processes allocates a console, which Windows hands to `WindowsTerminal.exe`. The result is a black terminal window that appears on screen and takes focus roughly every 5 to 6 seconds for as long as the user is working.

The visible window is Claude Code's own IDE-detection probe, not hivemind code. But hivemind is what causes it to run on a loop, so the two are only a problem in combination.

---

## Root cause: observed process ancestry

Captured with a WMI `__InstanceCreationEvent` subscription (`WITHIN 0.05`), resolving each new process's ancestor chain at creation time. Polling-based sampling misses these; they live well under 250ms.

```
node.exe   .../hivemind/0.7.131/bundle/wiki-worker.js  <tempfile deeplake-wiki-<session-id>>
  └─ cmd.exe  /d /s /c "…\npm\claude.cmd -p --no-session-persistence
                        --model haiku --permission-mode bypassPermissions"
       └─ claude.exe   (headless Claude Code)
            └─ cmd.exe /d /s /c "tasklist | findstr /I "Code.exe Cursor.exe Windsurf.exe
                                 Devin.exe idea64.exe pycharm64.exe webstorm64.exe
                                 phpstorm64.exe rubymine64.exe clion64.exe goland64.exe
                                 rider64.exe datagrip64.exe appcode.exe dataspell64.exe
                                 aqua64.exe gateway64.exe fleet.exe studio64.exe""
                 └─ tasklist.exe  +  findstr.exe
[...]

PR #332, opened 2026-07-29
fix(summary): back off after a failed summary run instead of respawning per event

efenocchi:
## Problem

Reported in #331: on Windows a headless `claude -p` spawns roughly every 5-6 seconds during normal tool work, flashing a console window, burning model API calls, and holding Defender at ~19.6% CPU. #256 hid the window; this fixes the spawning.

The reporter attributed it to "one spawn per capture event by design". It isn't — the periodic trigger is gated at 50 events / 2 hours (`summary-state.ts`), and the spawn lock is held for the duration of a run, so a healthy 15-60s summary run physically cannot re-spawn at that cadence.

It's a hot loop on the failure path:

- `wiki-worker.ts` swallows a failing `claude -p` and sets `execSucceeded = false`
- it then returns early, so **`finalizeSummary` is never called**
- its `finally` releases the spawn lock immediately
- `lastSummaryCount` is therefore still `0`, so `shouldTrigger`'s first-summary clause (`lastSummaryCount === 0 && totalCount >= FIRST_SUMMARY_AT`) stays true **forever**

Once any summary run fails, every subsequent captured event spawns a fresh `claude -p` for the rest of the session. One missing counter update produces all three symptoms.

## Fix

Track attempts separately from successes:

- `markSummaryAttempt` stamps `lastAttemptAt` + `attemptsSinceSuccess` before a run is spawned
- `finalizeSummary` resets the counter on success
- `shouldTrigger` holds off for an exponential window (1, 2, 4, 8, 16, capped at 30 minutes) while a failure is outstanding

Wired into all five periodic triggers: claude_code, codex, cursor, hermes, and the standalone pi extension.

A healthy session is unaffected — a successful run clears the counter before the next trigger is ever evaluated.

## Verification

The regression test replays the real loop (bump per event, spawn when `shouldTrigger` fires, never finalize) over 30 events. It fails against the unfixed code, which is the point:

```
× collapses a 30-event failing session from 21 spawns down to 1
  AssertionError: expected 21 to be 1     ← back-off removed
Tests  65 passed (65)                      ← back-off restored
```

Also covered: the back-off schedule, that a clean session is unaffected, that `markSummaryAttempt` does not advance `lastSummaryCount`, that `finalizeSummary` clears the counter, and that a throwing stamp releases the spawn lock.

`tsc --noEmit` clean. Three pre-existing failures in the full suite (`flush-memory` needs network, `cli-bundle-runtime` needs a built bundle, `cli-update` fails only under parallel load and passes 35/35 alone) reproduce identically on `main` without this branch.

## Also here

Documents `HIVEMIND_SUMMARY_EVERY_N_MSGS`, `HIVEMIND_SUMMARY_EVERY_HOURS`, `HIVEMIND_WIKI_WORKER` and `HIVEMIND_GRAPH_ON_STOP`, which all existed in the bundle but were undocumented — issue suggestion #4. Without them the only discoverable remedy was disabling the whole plugin.

## Review notes

Codex reviewed this branch and confirmed the root cause against the source. Two of its findings are fixed here (the lock leak when the stamp throws; the pi extension keeping the old behavior and stripping the new fields from shared state). Three it raised are pre-existing and deliberately not in scope:

- `SessionEnd` skips when a periodic worker holds the lock, so a final summary can still be lost if that worker then fails
- the 10-minute lock is a lease with no renewal or owner check, so a slow worker can be overtaken
- a successful upload followed by a failed `finalizeSummary` leaves the session retrying; with this change that is a capped 30-minute cycle instead of a per-event one

## Not verified on Windows

The mechanism is confirmed by reading the code and by the simulated-loop test, but nobody has run this on a real Windows box. Worth asking @jzferrell26 to re-test once this and #256 ship rather than closing #331 on our own say-so.

Refs #331

## Summary by CodeRabbit

- **New Features**
  - Added configuration options for periodic summary frequency, wiki worker activation, and code-graph rebuild behavior.
  - Added automatic retry backoff for unsuccessful periodic summaries, with delays increasing up to 30 minutes.

- **Bug Fixes**
  - Prevented repeated summary attempts from triggering on every captured event after a failure.
  - Ensured summary retry state resets after a successful completion.
  - Improved lock handling when recording or starting a summary attempt.

- **Tests**
  - Added coverage for retry timing, state persistence, trigger ordering, and failure scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,10 @@ This plugin captures session activity and stores it in your Deeplake workspace:
 | `HIVEMIND_CAPTURE`        | `true`                    | Set to `false` to disable capture          |
 | `HIVEMIND_CAPTURE_ONLY_CLI` | _(none)_                | Set to `true` to capture only interactive CLI sessions. Sessions spawned by the Claude Agent SDK (Python/TypeScript) are skipped; their `CLAUDE_CODE_ENTRYPOINT` is `sdk-py` / `sdk-ts`, so they fail the substring check for `cli`. |
 | `HIVEMIND_SKILLIFY_EVERY_N_TURNS` | `20`              | Assistant turns between auto skill-mining attempts. Lower = more frequent mining (cheaper sessions, noisier output); higher = fewer attempts on longer histories. |
+| `HIVEMIND_SUMMARY_EVERY_N_MSGS` | `50`                | Captured events between periodic session summaries. The first summary of a session runs at 10 events regardless. Raise it to cut background summary runs. |
+| `HIVEMIND_SUMMARY_EVERY_HOURS` | `2`                  | Time-based summary cadence, used when at least one new event has arrived since the last summary. |
+| `HIVEMIND_WIKI_WORKER`    | _(none)_                  | Set to `1` to disable the background session-summary worker entirely (no `claude -p` summary runs). Also set automatically inside the worker as a recursion guard. Capture and recall keep working. |
+| `HIVEMIND_GRAPH_ON_STOP`  | _(none)_                  | Set to `0` to disable the code-graph rebuild that runs on `Stop` / `SessionEnd`. |
 | `HIVEMIND_EMBEDDINGS`     | `true`                    | Set to `false` to force lexical-only mode  |
 | `HIVEMIND_PROACTIVE_RECALL_DISABLED` | _(none)_       | Set to `1` to disable **proactive recall** (auto-searching team memory on each recall-worthy prompt and injecting a relevant snippet into the agent's context). On by default. Does **not** affect capture or the agent's own grep/skill recall. Alt form: `HIVEMIND_PROACTIVE_RECALL=0`. |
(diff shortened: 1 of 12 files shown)


efenocchi:
@coderabbitai review

efenocchi:
Both findings fixed in c2ef660a.

**Major — fail closed on an unpersisted stamp.** `writeSummaryState` swallowed every write error, so a stamp that never reached disk still let the worker spawn, and the next captured event refired immediately — the exact behavior this PR exists to prevent. It now returns a boolean, and the periodic path releases the summary lock and skips the spawn when state can't be read or the stamp can't be persisted. Codex's review independently flagged the same thing ("pi's stamp is best-effort").

**Minor — the documented off switch.** You're right that the README over-promised. Rather than narrow the docs, pi now honors `HIVEMIND_WIKI_WORKER=1` in `maybeTriggerPeriodicSummary`, which doubles as the same recursion guard the bundled hooks use — so the table is accurate for every runtime.

Covered by three new assertions in `tests/pi/pi-extension-source.test.ts` (release-then-return ordering on stamp failure, the boolean return, the env guard).

On docstring coverage: not addressed. The 40% figure is dominated by pre-existing functions in the files this PR touches, and adding docstrings to unrelated functions would inflate the diff.

PR #347, opened 2026-09-03
fix: grant summarizer dirs explicitly instead of relying on bypassPermissions

efenocchi:
Fixes a reported field failure: on a host whose enterprise policy sets
`"disableBypassPermissionsMode": "disable"`, the session summarizer is silently
dead — every summary is a header-only stub and every backfill session reports
`no-summary`.

## Root cause

The wiki worker spawns a child `claude -p` with `--permission-mode bypassPermissions`
and then hands it absolute paths **outside the session cwd** (`$TMPDIR/deeplake-wiki-<id>-<ts>/`).
Under that policy the bypass is ignored, the child falls back to normal
permissioning, refuses every out-of-cwd path, and — because print mode cannot
prompt — exits 0 having written nothing:

```
running claude -p
claude -p exited (code 0)
no summary file generated
```

**The second bug is what makes it permanent.** When SessionStart has pre-seeded a
placeholder, the worker takes the SUCCESS path, uploads the placeholder verbatim,
and stamps `lastSummaryCount` — slicing those events away forever. Repeat every
run and every summary is a header stub with no body, by construction. The
`summaryChanged` guard was only consulted when the child exited **non-zero**.

## The fix

Name the dirs and tools instead of relying on the bypass. `--add-dir` +
`--allowedTools Read Write` is both policy-proof and least-privilege — the
summarizer never needs more than Read and Write.

- `src/hooks/wiki-worker-spawn.ts` — `CLAUDE_FLAGS` splits into `CLAUDE_BASE_FLAGS`
  + `CLAUDE_BYPASS_FLAGS`; a new `ClaudeGrants` interface and `permissionFlags()`
  helper. A caller that supplies grants gets them **instead of** the bypass; a
  caller that supplies none keeps the old blanket bypass, so nothing else changes.
  The Windows `.cmd` branch quotes granted paths, since a temp dir there routinely
  contains spaces.
- `src/hooks/wiki-worker.ts` — grants `tmpDir`, and applies the `summaryChanged`
  guard on the success path too.
- `src/skillify/stage-memory.ts` — grants the backfill's transcript dir and
  staging dir; `runAgent` grows an optional `grants` param.

## The other four `bypassPermissions` call sites

Audited all of them rather than shipping a partial fix.

| Call site | Hands the child an out-of-cwd path? | Action |
| --- | --- | --- |
| `src/commands/mine-local.ts` | Yes — the gate prompt names a verdict path under the per-session tmp dir | **Granted** |
| `src/skillify/gate-runner.ts` | Yes — same, via the skillify worker's tmp dir | **Granted** |
| `src/skillify/advisor.ts` | No — prompt is inlined, verdict is parsed from stdout | No change |
| `src/docs/refresh-llm.ts` | No — source content is inlined, markdown is read back from stdout | No change |

The two granted sites both offer the model a stdout fallback, so they degraded
rather than failed outright; under the policy the Write-tool branch was dead and
the verdict depended on the model choosing to print instead. `refresh-llm.ts`'s
codex branch uses `--dangerously-bypass-approvals-and-sandbox`, a different CLI's
flag, unaffected by the Claude enterprise policy. Non-claude agents keep their
existing argv untouched.

## Tests

The point of the regression test is the *silent success*, not the flags — a test
that only checked the argv would pass against a broken implementation.

- **`tests/claude-code/wiki-worker.test.ts`** — an exit-0 run that writes nothing
  must NOT upload the pre-seeded summary and must NOT advance `lastSummaryCount`.
  Confirmed to **fail** against the pre-fix worker (removing the guard turns it
  red), so it genuinely pins the bug. Its mirror asserts a run that *does* rewrite
  the summary still uploads and stamps, so the fix cannot silently freeze summaries.
- **`tests/shared/claude-permission-grants.test.ts`** (new) — grants replace the
  bypass; no grants keep it byte-identical; an empty grants object degrades to the
  bypass rather than granting nothing; the Windows `.cmd` branch quotes a path
  containing spaces.
- **`skillify-gate-runner` / `mine-local-orchestrator`** — argv assertions for both
  halves of the contract at the two newly granted call sites.

## What was and was not verified

- **Verified here:** the full suite is green (5898 tests) with per-file coverage
  thresholds met, `tsc --noEmit` clean, and the built argv carries the grants and
  not the bypass. The offset does not advance on an unchanged summary — proven by
  removing the guard and watching the test go red.
- **Verified out-of-band**, on a policy-managed macOS host this repo's CI cannot
  reproduce: the wiki prompt yields a real summary where it previously yielded
  none, and `stageSession` returns `ok:true` instead of `reason:"no-summary"`.
- **NOT verified from CI or from this machine:** behaviour under a real enterprise
  policy. `disableBypassPermissionsMode` is a managed-settings knob; unit tests
  cannot reproduce the environment, only the argv and the offset bookkeeping.

## Summary by CodeRabbit

* **Security**
  * Claude-powered tasks now receive access only to required directories and Read/Write tools.
  * Fallback permission behavior remains available when no specific permissions are provided.

* **Bug Fixes**
  * Wiki summaries are uploaded and session progress advances only after the summary is updated.
  * Legitimate rewrites with identical content are now recognized as updates, while summaries that were never written are skipped.

* **Tests**
  * Expanded coverage for permission handling, summary updates, supported platforms, and different agent types.

---

# Verification

## 1. The policy is actually honoured — not simulated

`claude` accepts `--settings`, so the child genuinely enforces
`disableBypassPermissionsMode: "disable"`: the bypass flag is *passed* and the CLI
*ignores* it, which is precisely the customer's condition. Full 2×2 with controls,
real `claude` 2.1.258, target dir outside the cwd:

| | old argv (`--permission-mode bypassPermissions`) | new argv (`--add-dir` + `--allowedTools Read Write`) |
| --- | --- | --- |
| **policy OFF** | summary WRITTEN (71 B) | summary WRITTEN (80 B) |
| **policy ON** | **exit 0, summary NOT WRITTEN** — the bug | summary WRITTEN (56 B) — the fix |

Exactly one cell fails and it is the reported one; both control cells pass, so the
failure is attributable to the policy and nothing else. In the failing run the child
states it itself: *"the write is blocked at the system level"* — the
`claude -p exited (code 0)` + `no summary file generated` pair from the field log.

## 2. End-to-end on the built worker: pre-fix vs fixed

Real `wiki-worker.js` bundle run as a real process against a local stand-in for the
query endpoint, isolated `HOME`, no live table touched. Same input, two bundles:

| Bundle | Uploaded a summary? | Advanced `lastSummaryCount`? |
| --- | --- | --- |
| pre-fix (today's `main`) | **YES** — `uploaded /summaries/…/sid-e2e.md (summary=69, desc=11)` | **YES** — `lastSummaryCount=9` |
| fixed | no — `exited 0 but left the pre-seeded summary unchanged; skipping upload` | no |

Those 69 bytes are the placeholder re-uploaded verbatim, and the stamped offset is the
moment the 9 unread events are lost for good.

With the **real** `claude` binary and a fresh session, the fixed worker produces a real
summary and uploads it (`INSERT INTO "memory"` observed), so the grants do not merely
block the damage — the working path still works.

## 3. The codex harness, with the real `codex` binary

| Case | Uploaded? | Offset |
| --- | --- | --- |
| agent exits 0 having written nothing (pre-seed present) | no — `codex exec exited 0 but left the pre-seeded summary unchanged` | not advanced |
| real `codex`, fresh session | YES — `uploaded … (summary=293, desc=9)` | `lastSummaryCount=9` |

Cursor, hermes and pi carry the identical guard and unit tests, but their CLIs are not
signed in on the verification host, so they are covered by tests + mutation only.

## Review consensus

Four adversarial review passes (codex, non-author). Each of the first three blocked, and
each block was a real defect — two of them regressions introduced by this branch while
fixing the original bug:

| Pass | Finding | Outcome |
| --- | --- | --- |
| 1 | Identical-content rewrite would freeze the offset forever | fixed |
| 1 | `commit-kpi-extract.ts:121` spawns `claude -p` with no grant | verified, documented below, out of scope |
| 1 | Two tests passed against a broken implementation | fixed, both verified to bite |
| 2 | mtime check had a hole on a coarse-resolution filesystem clock | fixed with a backdated sentinel |
| 3 | `utimesSync` outside the try — a rejecting filesystem aborted the worker | fixed |
| 3 | A silently-ignored `utimes` reopened the hole | fixed via `trusted: false` + content fallback |
| 4 | — | **VERDICT: SHIP** |

The wrote-nothing decision lived in five copies across the workers — which is how the
original bug propagated in the first place — so it now lives once in `wiki-offset.ts`
(`markSummaryUnwritten` / `summaryWasWritten`), which all five already import, with seven
direct unit tests including both `utimes` failure modes.

When the filesystem makes timestamps untrustworthy the check falls back to content
comparison, which errs toward **skipping** the upload: re-summarizing the same rows next
run wastes work, whereas a wrong upload destroys events. Never the other way round.

## Still not verified

- A **real** managed-settings policy file (`/etc/claude-code/managed-settings.json`,
  or the macOS path). `--settings` makes the CLI honour the same key and reproduces the
  reported behaviour exactly, but it is not the same settings source.
- cursor / hermes / pi against their real CLIs.

## Out of scope, found while verifying — worth a separate fix

`src/hooks/wiki-worker.ts:88` — `query()` retries only on HTTP *status* (401/403/429/5xx).
A network error **thrown** by `fetch` (dropped keep-alive socket, connection reset) is not
caught and kills the worker with `fatal: fetch failed`, losing the summary. The gap between
the pre-`claude` SELECTs and the post-`claude` upload is exactly the duration of
`claude -p` — tens of seconds — which is when an idle pooled socket is most likely to have
been closed. Same user-visible symptom as this bug, different cause. Not touched here.

### DIFF
--- a/src/commands/mine-local.ts
+++ b/src/commands/mine-local.ts
@@ -38,6 +38,7 @@ import {
 import { extractPairs, type Pair } from "../skillify/extractors/index.js";
 import { findAgentBin, type Agent } from "../skillify/gate-runner.js";
 import { extractJsonBlock } from "../skillify/gate-parser.js";
+import { permissionFlags, type ClaudeGrants } from "../hooks/wiki-worker-spawn.js";
 import { resolveSkillsRoot, writeNewSkill, listSkills, parseFrontmatter } from "../skillify/skill-writer.js";
 import { detectAgentSkillsRoots } from "../skillify/agent-roots.js";
 import { fanOutSymlinks } from "../skillify/pull.js";
@@ -91,6 +92,12 @@ function runGateViaStdin(opts: {
   bin: string;
   prompt: string;
   timeoutMs: number;
+  /**
+   * Dirs/tools to grant explicitly instead of the blanket `bypassPermissions`,
+   * which an enterprise policy can disable. See ClaudeGrants in
+   * ../hooks/wiki-worker-spawn.ts.
+   */
+  grants?: ClaudeGrants;
 }): Promise<{ stdout: string; stderr: string; errored: boolean; errorMessage?: string }> {
   return new Promise((resolve) => {
     if (opts.agent !== "claude_code") {
@@ -116,7 +123,7 @@ function runGateViaStdin(opts: {
       "-p",
       "--no-session-persistence",
       "--model", "haiku",
-      "--permission-mode", "bypassPermissions",
+      ...permissionFlags(opts.grants),
     ];
     const child = spawn(opts.bin, args, {
       stdio: ["pipe", "pipe", "pipe"],
@@ -586,7 +593,16 @@ async function runMineLocalImpl(args: string[]): Promise<void> {
     const prompt = buildSessionPrompt(tail, s, verdictPath);
     writeFileSync(join(sessionTmp, "prompt.txt"), prompt);
 
-    const gate = await runGateViaStdin({ agent: gateAgent, bin: gateBin, prompt, timeoutMs: GATE_TIMEOUT_MS });
+    // sessionTmp holds the verdict path named in the prompt and lives outside
+    // the session cwd, so grant it explicitly rather than relying on the
+    // bypass an enterprise policy can disable.
+    const gate = await runGateViaStdin({
+      agent: gateAgent,
+      bin: gateBin,
+      prompt,
+      timeoutMs: GATE_TIMEOUT_MS,
+      grants: { addDirs: [sessionTmp], allowedTools: ["Read", "Write"] },
+    });
     try {
       writeFileSync(join(sessionTmp, "gate-stdout.txt"), gate.stdout);
       if (gate.stderr) writeFileSync(join(sessionTmp, "gate-stderr.txt"), gate.stderr);
--- a/src/hooks/codex/wiki-worker.ts
+++ b/src/hooks/codex/wiki-worker.ts
@@ -13,7 +13,7 @@ import { buildTrailingPromptInvocation } from "../wiki-worker-spawn.js";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 import { finalizeSummary, releaseLock, readState } from "../summary-state.js";
-import { capLinesByBytes, newRowsFromWindow, stampOffset, WIKI_FALLBACK_MAX_ROWS, WIKI_JSONL_MAX_BYTES } from "../wiki-offset.js";
+import { capLinesByBytes, markSummaryUnwritten, newRowsFromWindow, stampOffset, summaryWasWritten, WIKI_FALLBACK_MAX_ROWS, WIKI_JSONL_MAX_BYTES } from "../wiki-offset.js";
 import { redactSecrets } from "../shared/redact.js";
 import { uploadSummary } from "../upload-summary.js";
 import { log as _log } from "../../utils/debug.js";
@@ -239,14 +239,7 @@ async function main(): Promise<void> {
     wlog("running codex exec");
     let execSucceeded = false;
     const summaryBeforeExec = existsSync(tmpSummary) ? readFileSync(tmpSummary, "utf-8") : null;
-    // Backdate the pre-seeded file before handing it to the child. Comparing
-    // mtime against "a minute ago" instead of "just now" closes the only hole
-    // in the wrote-nothing check: on a coarse-resolution filesystem (FAT's 2s,
-    // say) a same-tick rewrite would otherwise keep the timestamp and read as
-    // "never written". Any write by the child lands far outside that window.
-    const MTIME_SENTINEL = new Date(Date.now() - 60_000);
-    if (existsSync(tmpSummary)) utimesSync(tmpSummary, MTIME_SENTINEL, MTIME_SENTINEL);
-    const summaryMtimeBefore = existsSync(tmpSummary) ? statSync(tmpSummary).mtimeMs : 0;
+    const summaryBaseline = markSummaryUnwritten(tmpSummary, { existsSync, utimesSync, statSync });
     try {
       const inv = buildTrailingPromptInvocation(cfg.codexBin, [
         "exec",
@@ -284,11 +277,7 @@ async function main(): Promise<void> {
       // summary in place. Uploading it unchanged would still advance the offset
       // and slice those events away forever, which is how a session gets stuck
       // as a header-only placeholder run after run.
-      // mtime, not just content: an agent that legitimately regenerates the
-      // SAME text still touched the file, and skipping that would freeze the
-      // offset and re-summarize those rows on every future run. Only a run
-      // that neither changed the bytes NOR touched the file wrote nothing.
-      if (!summaryChanged && statSync(tmpSummary).mtimeMs === summaryMtimeBefore) {
+      if (!summaryWasWritten(tmpSummary, summaryBaseline, summaryChanged, { statSync })) {
         wlog("codex exec exited 0 but never wrote the summary; skipping upload to avoid advancing the offset");
         return;
       }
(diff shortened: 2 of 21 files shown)


efenocchi:
Addressed the review comment in b21b2e1f: both assertions now pin the granted directory value rather than just the flag name.

- `wiki-worker.test.ts` — asserts the argv tail is exactly `["--add-dir", tmpDir, "--allowedTools", "Read", "Write"]`.
- `mine-local-orchestrator.test.ts` — asserts the dir following `--add-dir` is this session's own tmp dir (the one holding the verdict path named in the prompt), followed by exactly `--allowedTools Read Write`.

Both were mutation-checked: swapping the granted dir for a neighbouring path turns each test red, so they now catch a grant that names the wrong directory — which would have left the child exactly as unable to reach the scratch dir as the bypass it replaces.

efenocchi:
Addressed in 22164255. Both comments were the same point and it is a good one for this PR specifically: `toContain("never wrote the summary")` is satisfied by all five workers, so a guard copy-pasted from one worker into another — precisely how the original bug reached all five — would have passed every one of those tests.

Each test now pins its worker's complete line (`claude -p` / `codex exec` / `cursor-agent --print` / `hermes -z` / `pi --print`). Verified by making the pi worker log codex's message: `tests/pi/pi-wiki-worker.test.ts` goes red.

efenocchi:
## Non-author review: APPROVED

Adversarial review by a separate agent (codex, non-author), four passes over this branch.
The first three each returned BLOCK, and each block was a real defect — two of them
regressions this branch introduced while fixing the original bug. All were fixed and
re-reviewed:

| Pass | Finding | Outcome |
| --- | --- | --- |
| 1 | Identical-content rewrite would freeze the offset forever | fixed (`0eddb2f8`) |
| 1 | Two tests passed against a broken implementation | fixed, both verified to bite |
| 2 | mtime check had a hole on a coarse-resolution filesystem clock | fixed (`a9cce846`) |
| 3 | `utimesSync` outside the try — a rejecting filesystem aborted the worker | fixed (`0ad55a03`) |
| 3 | A silently-ignored `utimes` reopened the hole | fixed via `trusted: false` + content fallback |
| 4 | — | **APPROVED** |

Final pass, verbatim:

> - Both blockers are closed. `wiki-offset.ts:163-170`: `utimesSync` and verification are inside `try`; failures return `trusted: false`.
> - `wiki-offset.ts:165-167`: the observed mtime is compared with the sentinel, detecting silently ignored stamping.
> - `wiki-offset.ts:179`: `if (!baseline.trusted) return contentChanged` safely rejects unchanged output, preventing offset advancement and event loss; legitimate identical rewrites may only be retried.
> - All workers invoke the shared helpers: `wiki-worker.ts:327,371`; `codex/wiki-worker.ts:242,280`; `cursor/wiki-worker.ts:277,321`; `hermes/wiki-worker.ts:279,333`; `pi/wiki-worker.ts:227,272`.
> - `wiki-summary-written.test.ts:61-86` covers throwing and silently ignored `utimes`; all 7 focused tests and typecheck pass.
> - No new issue found in the requested scope.
>
> **VERDICT: SHIP**

Two findings from the same review were deliberately left out of this PR rather than
silently folded in, and are recorded in the PR body: an exit-0 *partial* write is still
treated as success (pre-existing; this branch only narrows the exposure), and
`commit-kpi-extract.ts:121` spawns `claude -p` with no grant (currently dead code —
commit-driven KPI extraction is disabled at `capture.ts:236-241`).

CodeRabbit raised three comments across two reviews; all three were applied and each was
verified to actually catch the defect it describes.

Follow-up filed: PLA-500 — `query()` does not retry a thrown network error, killing the
summary run after the LLM call is already paid for.

efenocchi:
@coderabbitai All three comments from your previous two reviews have been addressed (b21b2e1f, 22164255), and each fix was verified to actually catch the defect it describes rather than just silence the comment. The branch has since had four adversarial review passes from a separate non-author agent; its three blocking findings are fixed in 0eddb2f8, a9cce846 and 0ad55a03.

Could you post your verdict on the current head (22164255)?

Issue #337, opened 2026-08-14
docs: README claims "BM25 fallback when embeddings off" — BM25 was tried and dropped


kaghni:
`README.md` describes retrieval as:

> Searches traces and skills with hybrid lexical + semantic retrieval (**BM25 fallback when embeddings off**)

That is not what the code does.

**BM25 was tried and removed** — `src/shell/grep-core.ts:351-355`:
> ```
> // BM25 tried and dropped (PR-NOTES F4c): score scale (~1..3) overpowered
> // cosine in UNION, semantic hits were pushed out of top-K. LIKE is a
> // better fit for "find any session mentioning X" which is the actual
> // plugin use case.
> ```

**BM25 index creation is also disabled** — `src/deeplake-api.ts:548-549`:
> ```
> // BM25 index disabled — CREATE INDEX causes intermittent oid errors on fresh tables.
> // See bm25-oid-bug.sh for reproduction. Re-enable once Deeplake fixes the oid invalidation.
> ```

The actual hybrid (`grep-core.ts:343-350`) is a single `UNION ALL` of lexical `LIKE`/`ILIKE` rows emitting a **sentinel score of 1.0** (capped by `HIVEMIND_HYBRID_LEXICAL_LIMIT`, default 20) with semantic rows emitting real cosine, ordered by score — so exact substring matches always outrank semantic ones.

`src/embeddings/disable.ts:26` carries the same stale wording ("falls back to BM25 / ILIKE matching").

**Suggested fix:** update the README bullet and the `disable.ts` comment to say lexical `LIKE`/`ILIKE`, and describe the sentinel-score fusion. Worth stating the rationale too — it's a good design note: BM25's unbounded score scale is incompatible with cosine in a single `ORDER BY` without rank-based fusion (RRF) or normalization.

PR #348, opened 2026-09-08
Fix: hybrid lexical + semantic retrieval ordering

shivansh31414:
## Summary

Fix stale documentation that described the hybrid retrieval strategy as using **BM25 as a fallback**. BM25 was evaluated and removed (PR-NOTES F4c) because its unbounded score scale (~1–3) overwhelms cosine similarity values (0–1) in a shared `ORDER BY`, causing semantic hits to be pushed out of the top-K entirely. Using it correctly would require rank-based fusion (RRF) or score normalisation, which adds complexity without clear benefit for the "find any session mentioning X" use case.

The actual hybrid strategy is a `UNION ALL` of:
- **Lexical** `LIKE`/`ILIKE` substring rows, emitting a **sentinel score of 1.0** (capped by `HIVEMIND_HYBRID_LEXICAL_LIMIT`, default 20)
- **Semantic** cosine-similarity rows, emitting their **real 0–1 score**

Results are ordered by score descending, so exact keyword matches always lead the ranking while semantic concept matches fill in below. When embeddings are off, retrieval falls back to lexical `LIKE`/`ILIKE` only.

**Files changed:**
- `README.md` — updated the feature bullet to describe the sentinel-score fusion and document why BM25 was dropped
- `src/embeddings/disable.ts` — corrected the JSDoc comment from `"BM25 / ILIKE matching"` to `"lexical LIKE/ILIKE matching"` with the same rationale

## Version Bump

> **To trigger a release**, bump `"version"` in `package.json` before merging.
>
> | Change type     | Version bump          | Example              |
> | --------------- | --------------------- | -------------------- |
> | Bug fix         | patch (1.2.0 → 1.2.1) | `"version": "1.2.1"` |
> | New feature     | minor (1.2.0 → 1.3.0) | `"version": "1.3.0"` |
> | Breaking change | major (1.2.0 → 2.0.0) | `"version": "2.0.0"` |
>
> If you don't bump the version, no release will be created.

This is a documentation-only change — no release needed. No version bump required.

## Test plan

- [ ] Tests pass locally (`npm test`)
- [ ] Relevant new tests added
- [x] Version bumped in `package.json`, or **no release needed for this change** ✓ (docs only)

## Summary by CodeRabbit

- **Documentation**
  - Clarified the search documentation’s description of lexical and semantic result scoring and ordering.
  - Documented that lexical-only matching uses `LIKE`/`ILIKE` when embeddings are disabled.
  - Added context explaining that BM25 is not used because its score scale is incompatible with cosine similarity without normalization or score fusion.


efenocchi:
Closing in favour of #350, which fixes the same stale BM25 wording from #337 with a minimal change to the three affected spots (README, docs/EMBEDDINGS.md, disable.ts JSDoc) and has a green CI run.

Your reading of the hybrid ranking is correct (UNION ALL, 1.0 sentinel for lexical rows, ORDER BY score, BM25 dropped per PR-NOTES F4c), but that explanation already lives in src/shell/grep-core.ts and is too much detail for the README feature list. Two small things here also needed fixing before it could land: the JSDoc in src/embeddings/disable.ts was left with a dangling "readable." line, and the PR title says "Fix" for a docs-only change. Thanks for taking the time to verify it against the code.

PR #359, opened 2026-09-17
fix(cowork-ingest): redact secrets before queuing transcript rows (#308)

### LINKED ISSUE #308: Redact secrets on the Cowork transcript ingest path (+ tests)
Follow-up to #307 (secret redaction on capture).

#307 masks secrets before persist/embed on the four agent capturers + `uploadSummary`. The **Cowork transcript ingest** path (`src/mcp/cowork-ingest.ts`, `entriesForLine` → queued session rows) is NOT yet redacted — deferred because that file is ~35% covered and wiring redaction into #307 dragged the PR's coverage comment down.

**Scope:**
- Apply `redactSecrets(...)` to the serialized entry in `cowork-ingest.ts` (the `const serialized = JSON.stringify(entry)` site), matching the pattern in the agent capturers.
- Add a real test suite for `cowork-ingest.ts` (currently ~35%): cover `entriesForLine` (pure transcript→entries mapping) and assert a secret in a Cowork transcript entry is masked before it's queued/embedded.

Until this lands, secrets echoed in Cowork transcripts are captured verbatim.

sumitvairagar:
## What

Closes #308.

Cowork transcript entries were serialized with `JSON.stringify(entry)` and passed directly to `buildQueuedSessionRow`. Secrets echoed in a Cowork session — API keys, Bearer tokens, passwords — reached the Deeplake store unmasked.

## Why it was missed

The four agent capturers (`capture.ts`, `codex/capture.ts`, `cursor/capture.ts`, `hermes/capture.ts`) all apply `redactSecrets()` at their serialization site. The Cowork ingest path was deferred from PR #307 because `cowork-ingest.ts` had lower coverage and wiring redaction in at that point dragged down the coverage comment.

## Fix

One-line change at the `buildQueuedSessionRow` call site in `ingestCoworkSessions()`:

```ts
// before
line: JSON.stringify(entry),

// after
line: redactSecrets(JSON.stringify(entry)),
```

Redacting the serialized line covers every field — `content`, `tool_input`, `tool_response` — in one pass. Identical pattern to every other capturer.

## Tests

Added 4 cases to `tests/claude-code/cowork-ingest.test.ts`:

| Case | Secret type | Field |
|------|-------------|-------|
| OpenAI `sk-` key in a user message | known token scheme | `content` |
| GitHub PAT in a `curl` Authorization header | known token scheme | `tool_input` |
| Anthropic `sk-ant-` key in a tool response | known token scheme | `tool_response` |
| Plain non-secret message | — | `content` (unchanged) |

Secrets are assembled from split literals at runtime so the source file does not contain scannable vendor tokens (matching the pattern in `tests/shared/redact.test.ts`).

18/18 `cowork-ingest` tests pass. `tsc --noEmit` clean. Pre-existing failures in `cowork-queue-leak` and `install-cowork` are unrelated and unchanged.

## Summary by CodeRabbit

* **Security**
  * Secrets in Cowork session messages, tool inputs, and tool responses are automatically redacted before processing.
  * Sensitive API keys and access tokens are masked while preserving safe, recognizable prefixes.

* **Reliability**
  * Cowork session ingestion now handles missing session identifiers and timestamps safely.

* **Tests**
  * Added coverage for secret masking across user messages, tool calls, and tool results.
  * Confirmed that non-sensitive content remains unchanged.


sumitvairagar:
Thanks for the review — both points are valid.

Working on fixes:
1. Extract a source-owned mapping function from `cowork-ingest.ts` so the tests exercise the production path rather than duplicating the redaction logic.
2. Tighten all assertions to exact expected values instead of substring checks.

Will push shortly.

efenocchi:
## Independent review + real-flow verification (non-author)

**Real flow.** The built MCP server from this branch (`mcp/bundle/server.js`, what Cowork spawns) started over an isolated `HOME` with a Cowork transcript at `<claudeDesktopConfigDir>/local-agent-mode-sessions/.../.claude/projects/<enc>/<sid>.jsonl` carrying an Anthropic key, a GitHub token, an AWS key id, a `password=` and an `AWS_SECRET_ACCESS_KEY=` tool result; rows read back from the staging table `sessions_e2e_removals`:

- main build (`3fdfd6e9`): all five secrets stored in plaintext — #308 confirmed.
- this branch (`5446cc68`): `user_message` → `export ANTHROPIC_API_KEY=******** and use password=******** for the db`; `assistant_message` → `Using token ghp_******** and aws key AKIA******** now.`; `tool_result` → masked, but stored as a `raw_message` blob because `redactSecrets` swallowed the JSON escape after the value. That is a pre-existing bug in `src/hooks/shared/redact.ts` that hits every capturer using `redactSecrets(JSON.stringify(entry))` (`src/hooks/capture.ts:177`); fixed separately in #360. With #360 applied the third row is a structured `tool_result` with `"tool_response":"\"AWS_SECRET_ACCESS_KEY=********\""`.

**Code review.** Redaction happens on the serialized line before `buildQueuedSessionRow`, parity with `capture.ts:177`; single call site (`cowork-ingest.ts:486`); `session_id ?? ""` / `timestamp ?? now` are unreachable defaults since `entriesForLine` always sets both; tests call the production `buildCoworkQueueRow`. `tsc` clean, `cowork-ingest.test.ts` 18/18.

VERDICT: APPROVED

